PR #19, opened 2012-09-18
Fixed long delay bug and minor doc improvements

xenomachina:
Hello. I made a few minor improvements to Arduino-IRremote.

The delay thing, in case you're wondering, came up because [I was trying to control a Lutron GrafikEye lighting controller](http://www.flickr.com/photos/laurence/7966467276/in/photostream), and one of the "marks" it expects is about 20ms, which is way longer than Arduino's delayMicroseconds is documented to support (and actually, I only learned this after witnessing my Diecimila "misbehave").

The naming/doc changes were in response to things that tripped me up, or forced me to carefully read the implementation.

```
- Delays longer than 16383us were unreliable, due to a quirk/bug in
  Arduino's delayMicroseconds.

- Some parameters that were meant to hold kHz values were named "hz", so
  renamed them to "khz".

- The meaning of sendRaw's buf parameter wasn't obvious to me, so I
  added a comment explaining it.
```

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -105,9 +105,11 @@ void IRsend::sendSony(unsigned long data, int nbits) {
   }
 }
 
-void IRsend::sendRaw(unsigned int buf[], int len, int hz)
+// buf is alternating mark and space durations (in microseconds), starting with
+// a mark at index 0.
+void IRsend::sendRaw(unsigned int buf[], int len, int khz)
 {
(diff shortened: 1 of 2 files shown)


z3t0:
@xenomachina Very much appreciated! Will rebase and commit eventually...

z3t0:
Commit 1e72064557cd implements a custom delay

PR #27, opened 2013-04-11
Add a decoder function pointer to IRrecv to help specify decoder.

aguegu:
The IRrecv::decode() would run through all decode functions until it gets what we want. But when it comes to specific application, when the remote controller is fixed. There would be no more need to run all decoders all over again and again. it would run faster. And in this way, receiver could avoid interference from other controllers. 

I have apply auto-format a few times in eclipse cdt to help my editing. So it would feel like a lot of commits. But it is not. Just add IRrecv::setDecoder() and update the IRrecvDemo example.

The NEC_ONE_SPACE is modified to 1680, as it should NEC_BIT_MARK * 3.


z3t0:
@aguegu Seems like a useful change, will be implemented during my OOB rewrite. I'll create an issue in place of this Pull Request

PR #32, opened 2013-07-10
Add overload to sendRC5 to support extended messages

Arachnid:
The RC5 protocol supports using the second start bit as a field selector, doubling the number of available device codes. This patch adds an overload to sendRC5 that allows users to specify the value of the field (second start) bit.


z3t0:
If anyone is interested feel free to rebase this off master, I will eventually get to it...

PR #33, opened 2013-07-13
Add IRrecv::decode which accepts only one encoding.

mslusarz:
As I explained in the commit: it can be used to reduce library overhead when only one type encoding is used - all code handling other encodings is optimized out.


z3t0:
#107 essentially solves the problem being addressed. Re-open if you disagree

z3t0:
#186 
Feel free to re-open/comment

PR #40, opened 2013-09-10
doubled RAWBUF size to handle my A/C remote

itamaro:
When decoding signals from the remote of my A/C (I have no idea what vendor..), I had to increase the size of RAWBUF, since I dealt with signals that needed 182-188 bins instead of the original 100.


z3t0:
#51

PR #47, opened 2013-12-16
Added ability to use active-high IR receivers with 2nd (optional) argument to IRrecv().  Still defaults to active-low.

agprimatic:
If one argument is provided to IRrecv(), then the input defaults to active-low.
If two arguments are provided to IRrecv(), the 2nd argument is the sense of the IR receiver (0 = active-low, 1 = active-high).

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -267,6 +267,14 @@ IRrecv::IRrecv(int recvpin)
 {
   irparams.recvpin = recvpin;
   irparams.blinkflag = 0;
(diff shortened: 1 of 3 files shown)


bengtmartensson (commented):
All commercially available demodulating IR receivers are active low.  The PR increases complexity, size and (possibly) execution time. This is a feature (almost?) nobody needs.

Also, it is not compatible with the current code base.

Recommendation: no merge.

ArminJo:
Thanks for the advise.

PR #48, opened 2013-12-21
add isIdle() method

StefanScherer:
Added isIdle() method to check if interrupt code is "receiving" a new IR
code.

With this information Wait until IRrecv is idle before starting another long running time critical
code with noInterrupts() inside like the NeoPixel code has to do with long LED strips.

    void show() {
      while (!irrecv.isIdle()) {
        delay(1);
      }
      strip.show();
    }

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -375,6 +375,10 @@ ISR(TIMER_INTR_NAME)
   }
 }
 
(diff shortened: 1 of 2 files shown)


z3t0:
Merged. 1e519b1

PR #51, opened 2014-02-06
Enabled raw code length >254

nitsujri:
AC units sometimes have large raw code lengths.

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -108,8 +108,8 @@ public:
 
 // Some useful constants
 
-#define USECPERTICK 50  // microseconds per clock interrupt tick
-#define RAWBUF 100 // Length of raw duration buffer
+#define USECPERTICK 15  // microseconds per clock interrupt tick
+#define RAWBUF 350 // Length of raw duration buffer
 
 // Marks tend to be 100us too long, and spaces 100us too short
 // when received due to sensor lag.
--- a/IRremoteInt.h
+++ b/IRremoteInt.h
@@ -195,7 +195,7 @@ typedef struct {
   uint8_t blinkflag;         // TRUE to enable blinking of pin 13 on IR processing
   unsigned int timer;     // state timer, counts 50uS ticks.
   unsigned int rawbuf[RAWBUF]; // raw data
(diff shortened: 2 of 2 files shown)


wiredmax:
What is the status on this merge?

z3t0:
everything is waiting on #139 unless

On Sat, May 30, 2015 at 1:10 PM, Maxime Poulin <notifications@github.com>
wrote:

> What is the status on this merge?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/shirriff/Arduino-IRremote/pull/51#issuecomment-107076527>
> .
>



-- 
Thanks,
Rafi Khan

z3t0:
#139 is essentially done so will take a look at merging this soon

z3t0:
@AnalysisIR any thoughts?

AnalysIR:
To be able to handle long AC signals, setting RAWBUF to 256 will still miss most of them.
(I believe RAWBUF is still limited to a max length of 256)

However, it would be useful if users could at least (dynamically) set RAWBUF length from their own code without having to dig into & modify the library files.

As a workaround we have offered an alternative sketch via our blog which is extremely popular and which handles long AC signals on most standard arduinos. Feel free to link to that as a standalone sketch, for long AC signals.
http://www.analysir.com/blog/2014/03/19/air-conditioners-problems-recording-long-infrared-remote-control-signals-arduino/

If you want to include it with the library as a standalone sketch/example/utility, just be sure to provide a link back to the original article on our site, with the usual credits etc.

(PS: I don't foresee a situation (in the sort term) where you will have a generic library for recording & decoding multiple AC signals on standard Arduinos. The best/likely target is to be able to record RAW AC signals.)

z3t0:
> However, it would be useful if users could at least (dynamically) set RAWBUF length from their own code without having to dig into & modify the library files.

What kind of sizes for `RAWBUF

AnalysIR:
As  per the blog 800+ which means standalone only

z3t0:
Closing this Pull Request. Will open an issue instead and refer it to @AnalysIR Sketch as it will be the preferred way to implement such communications until the library has a built in solution.

PR #60, opened 2014-04-24
WIP: Add support for Fujitsu heat pumps

GeorgeDewar:
I have added support to this library for my Fujitsu ASTG12LVCC air conditioner (with AR-RAH1E remote). My suspicion is that the code written so far is generic to many Fujitsu products.

Basically, the code is capable of recognizing and decoding a Fujitsu IR packet and making available the underlying (usually 128 bit) data. It can also send.

Here is an example (not quite working perfectly) of an Arduino application using this functionality to decode data (temperature, mode, fan speed) from such a device, and resend the packet.

This is a bit of a work in progress - I have figured out how the checksum is calculated and might want to add that to this.

I hope this work may be helpful to someone else who wants to (perhaps) control their AC unit with an Arduino!

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -914,37 +914,20 @@ long IRrecv::decodeFujitsu(decode_results *results) {
         return ERR;
     }
 
-    int temp = 0;
-
-    for(int i=0; i<4; i++) results->data[i] = 0;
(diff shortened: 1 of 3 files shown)


z3t0:
Please re-open if you have been successful, I would prefer not to merge unstable code that I have no way to test. Thanks for your contribution and efforts!

PR #90, opened 2014-09-25
PlatformIO Library Manager manifest file

ivankravets:
* This library in Web Registry: http://platformio.org/#!/lib/show/4/IRremote
* Specification: [PlatformIO Library Manager](http://docs.platformio.ikravets.com/en/latest/librarymanager/index.html)
* Integration: [IDE Integration](http://docs.platformio.ikravets.com/en/latest/ide.html)


z3t0:
Should this be merged now?

PR #92, opened 2014-10-14
Increased RawBuffer size

lucagiacalone85:
I've had many problems to record the IR signal sent from my Fujitsu air cooler remote controller. I've found an issue regarding the length of the buffer. My fujitsu sent a 260 byte legth signat to turn on the air cooler. So I've increase the length of the buffer up to 400. I hope this can help.

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -119,7 +119,7 @@ public:
 // Some useful constants
 
 #define USECPERTICK 50  // microseconds per clock interrupt tick
-#define RAWBUF 100 // Length of raw duration buffer
+#define RAWBUF 400 // Length of raw duration buffer
 
 // Marks tend to be 100us too long, and spaces 100us too short
 // when received due to sensor lag.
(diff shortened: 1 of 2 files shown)


z3t0:
@lucagiacalone85 Is it fine if I add this to the wiki instead of changing the code itself. There seems to be quite a lot of variations discovered by changing the RAWBUF .

lucagiacalone:
Ok but keep in mind that this kind of issue it's a little bit tricky. Probably many people can't undersand why the library doesn't work for some devices (like my air cooler). I personally spent 3 days to sort out this problem (actually I'm not an arduino expert). I think you should add a suggestion or a workaround to readme file to help people with same issue. 

BUT Thank you! the library is great! :D

z3t0:
OK that's a great idea. The website for this plugin is the perfect area for these kinds of tutorials. Would you mind documenting some of the steps you took in figuring out how to get your air cooler working. Thanks!

joshuajnoble:
I suggest this be closed with a note on the wiki that the buffer size can be increased but problems can ensue.

z3t0:
My bad.  I thought I had closed it.

PR #99, opened 2014-11-16
Allow timer selection via #define

tomdeboer:
This allows you to select which timer you want to use without touching IRremoteInt.h

You can `#define IR_USE_TIMER1` for example, to use Timer1 instead of the default (on Arduino) Timer2

### DIFF
--- a/IRremoteInt.h
+++ b/IRremoteInt.h
@@ -29,6 +29,7 @@
 // to switch IRremote to use a different timer.
 
 // Arduino Mega
+#if !defined(IR_USE_TIMER1) && !defined(IR_USE_TIMER2) && !defined(IR_USE_TIMER3) && !defined(IR_USE_TIMER4) && !defined(IR_USE_TIMER4HS) && !defined(IR_USE_TIMER5)
 #if defined(__AVR_ATmega1280__) || defined(__AVR_ATmega2560__)
   //#define IR_USE_TIMER1   // tx = pin 11
   #define IR_USE_TIMER2     // tx = pin 9
@@ -70,7 +71,7 @@
   //#define IR_USE_TIMER1   // tx = pin 9
   #define IR_USE_TIMER2     // tx = pin 3
 #endif
-
+#endif
 
 
 #ifdef F_CPU


z3t0:
@tomdeboer has this been tested?

tomdeboer:
No I'm afraid not.. It did work, but not in the Arduino IDE, only in Sublime or build file. Maybe you can give it another try.

z3t0:
OK I'll see if I can get it working

z3t0:
Does this work for all arduinos or only mega.  What would be a test sketch for an arduino uno?

If we get this working we may unlock support for alot more boards

z3t0:
What errors did you get when using the Arduino IDE

tomdeboer:
The problem with the Arduino IDE is that if you ```#define IR_USE_TIMER4``` in your .ino, the ```#if IR_USE_TIMER4``` would still evaluate false. I'm not sure, but I think the build file the IDE uses includes all the libraries includes all headers before the actual ```#include``` in your script. At that point your define isn't setup yet.

z3t0:
hmm

z3t0:
Just found this 
'''
The Arduino pre-processor doesn't understand #ifdef's or other pre-processor directives.  It links in libraries placed within "#if 0" and other false checks.  It also places the automatically generated function prototypes without regard for these directives, which often leads to various problems.  

This is tricky to fix and to test (there are lots of different cases).  For now, you may want to avoid the use of #ifdefs and #if 0 in your sketch or, at least, for commenting out unused libraries or at the beginning of your sketch (using them later in the code shouldn't affect the placement of the function prototypes).  
'''

z3t0:
"The #ifdefs work fine in .cpp, .c, and .h files, since those aren't pre-processed (that includes files in libraries, like in the tutorial you reference).  They work in your sketch files (.ino files) too but can interact with the Arduino pre-processor in ways that lead to the errors people have seen.  " - [here](https://code.google.com/p/arduino/issues/detail?id=906)

z3t0:
@tomdeboer How would we test this? To check if it actually works

tomdeboer:
Maybe you can create a separate .h and .cpp file with a setup and loop? In which you define your timer and include the IRremote.h (in that order). Because as I read that post correctly, .cpp files aren't processed by Arduino's pre-processor.

z3t0:
I'll see what I can do

z3t0:
Is there any way to test if it is using the timer we specify?

tomdeboer:
Maybe you can just use something like ```#warning "No time specified"``` if the #ifdef evaluates false.

z3t0:
@tomdeboer Would it make more sense to just create a page on the wiki explaining how to do this manually from within the library? I think that would be the simplest approach, see https://github.com/khair-eddine/Arduino-IRremote/pull/1/files

z3t0:
This is being closed as the approach does not work with the Arduino IDE.

PR #105, opened 2015-02-07
Added Whynter A/C remote protocol

fmeschia:
Tested with Whynter ARC-110WD


z3t0:
Has the author abandoned the project? If so I think we should fork it into a more updated version and all work on that.

shirriff:
No, I haven't abandoned it, but I don't have much time to put into it.

PR #107, opened 2015-03-09
Experimental

z3t0:
Added Sharp Decoding Support.
Please test and report results if you want this to become a feature and want this plugin to continue in its development.

PR #109, opened 2015-03-10
Debug Improvement, saving a lot of precious ram in debug mode

fab672000:
Saved a lot of ram in debug mode by switching the constant strings to the flash memory space, see the F() macro changes.
Minor Improvement:
  Saved one byte of ram per class instantiation as decode_type attribute only needs to be int8_t so 8bits

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -17,7 +17,6 @@
  * LG added by Darryl Smith (based on the JVC protocol)
  * Whynter A/C ARC-110WD added by Francesco Meschia
  */
-
 #include "IRremote.h"
 #include "IRremoteInt.h"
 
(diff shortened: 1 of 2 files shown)


IRremote.cpp
```diff
@@ -311,6 +310,74 @@ void IRsend::enableIROut(int khz) {
...
+    if (data & DISH_TOP_BIT) {
+      mark(DISH_BIT_MARK);
+      space(DISH_ONE_SPACE);
+    }
+    else {
+      mark(DISH_BIT_MARK);
+      space(DISH_ZERO_SPACE);
+    }
+    data <<= 1;
+  }
```
z3t0: Are these lines similar to the lines deleted below or rewritten?

IRremote.h
```diff
@@ -28,13 +35,15 @@
 // Results returned from the decoder
 class decode_results {
 public:
-  int decode_type; // NEC, SONY, RC5, UNKNOWN
+  int8_t decode_type; 	// NEC, SONY, RC5, UNKNOWN
```
z3t0: Nice catch!

IRremote.cpp
fab672000: Hi Rafi,
 yes they are just refactored so that all send code implementation can be conveniently ifdef'd and localized, no source modification at all here for IRsend.-Fab

z3t0:
OK thanks. I'll take another look and merge it once I get a chance. Thanks for the contribution. Mind if I add you to our contributors list on the [page](http://shirriff.github.io/Arduino-IRremote)?

fab672000:
NP Rafi,  glad I could help !-Fab

z3t0:
Just added your name, should appear soon.

z3t0:
Thank you very much for your contribution! Just merged. Please check #107 and test there. Report results there so that we can verify things are stable and move this into the master branch and eventually the next release.

PR #110, opened 2015-03-12
Added possibility to exclude non-necessary libraries to reduce code size

sstefanov:


z3t0:
Thanks! I'll test and report back results.

z3t0:
@sstefanov Does this produce measurable changes? I am interested because I remember closing an issue for a board with very limited space, so this could open possibilities for further support.

joshuajnoble:
I like this and unless @shirriff or @PaulStoffregen objects would like to merge it since there are so many different IR protocols and the code bloat to support all of them would make the library cumbersome to use with other libs on boards with limited space.

sstefanov:
When only one receiving protocol is enabled hex code is around 4 k less
than case when all send and receiving functions are enabled.
На 14.03.2015 00:32 "Rafi Khan" <notifications@github.com> написа:

z3t0:
@joshuajnoble Agreed. Should we merge this into the experimental branch then and add this feature to the new decoding that has already been added there?

z3t0:
@sstefanov Thanks for the specific details and also the work, that sounds like a huge prospect for those boards with limited space.

sstefanov:
I just made some tests and there are results (code is for remote controlled player). Size is for all code, not only for IR remote library:

All sending and receiving functions enabled:
Binary sketch size: 21314 bytes (of a 30720 byte maximum, 69.38 percent).

Only NEC receiving enbled:
Binary sketch size: 17918 bytes (of a 30720 byte maximum, 58.33 percent).

Only RC5 enabled:
Binary sketch size: 18926 bytes (of a 30720 byte maximum, 61.61 percent).

z3t0:
Awesome improvements. This will be merged as soon as @joshuajnoble and myself decide on the best approach to model a stable development system. Checkout #107 to get in on the conversation.

crash7:
Hi all,
Would be better to split the file in nec, rc5, sharp, etc. instead of using #ifdef everywhere? 
What do you think?

z3t0:
@crash7 sure we could do that.

joshuajnoble:
That makes more sense from a code perspective but it breaks the Arduino convention of having a single .h file to import the whole library, thus enabling the "Import Library" tool to just work.

Obligatory phone disclaimer

crash7:
@joshuajnoble true, nevermind then.

z3t0:
Oh sorry I misread, I thought you meant into separate PRs.

PR #113, opened 2015-03-15
cleaned up merge of #110

joshuajnoble:
This is actually #110 but it required some massaging to merge because of updates. Added a description to the README. I also thought that it makes more sense to have just #define JVC rather than requiring both #define JVC *and* #define IRsendJVC. 

Please ignore all the attiny related commits, they aren't included in this at all.


crash7:
@joshuajnoble It would be better to rebase this branch and only pick the commits related with #110, to avoid mixing those changes with the attiny ones.

What do you think?

joshuajnoble:
Agreed. I'll do that as soon as I get time.

Obligatory phone disclaimer

joshuajnoble:
Hmm. Ok, I honestly can't figure out how to get those out of my commit history. I think I have push privilege to shiriff/master so if it's important then I'll just do it that way.

PR #116, opened 2015-03-21
Updated #define directives for safely removal of the unused protocols.

vzhikserg:
I faced with a problem when I tried to comment out some protocols using the #define directive. The project didn't compile so I decided to update the source code. Some curly brackets were moved in order to apply the same style to the whole file.


z3t0:
No longer necessary, #107

PR #119, opened 2015-04-04
Fixed problem with interrupt enable/disable on ATmega8

jan-r:
Previously, when enabling or disabling interrupts on the ATmega8, the
whole TIMSK register was overwritten. This disables all other timer
interrupts (for Timer0 and Timer2). The fix takes care of that by
selectively enabling/disabling the required OCIE1A flag.


z3t0:
@jan-r Thank you for the contribution!

PR #124, opened 2015-04-06
Added different ifdefs for SEND and DECODE. Enum for decode_type. ifdef errors fixed.

levsa:


joshuajnoble:
I'd like to merge this now as it fixes the errors that were introduced in #110

z3t0:
Nice

PR #159, opened 2015-06-20
Re-Enable Interrupts on Resume

klasse85:
I've been using this library in a project where I use the same arduino for both sending and receiving IR-signals. This commit updates the library so that the interrupts will be re-enabled when calling resume (they are disabled when sending IR-signals).

Sorry for the "big" diff in IRremoteInt.h. Aside from adding the macro IS_TIMER_INTR_ENABLED, I've also added some white space to align the macros.

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -443,6 +443,13 @@ ISR(TIMER_INTR_NAME)
 }
 
 void IRrecv::resume() {
+if (!IS_TIMER_INTR_ENABLED()) {
+    // Enable the Timer2 Interrupt (which is used for receiving IR)
+    TIMER_CONFIG_NORMAL();
+    TIMER_ENABLE_INTR;
+    TIMER_RESET;
+    sei();
+  }
   irparams.rcvstate = STATE_IDLE;
   irparams.rawlen = 0;
 }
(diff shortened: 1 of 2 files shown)


AnalysIR:
**Caution!** - This may not be a good idea.

Many IR receivers can introduce a lag of over 200 uSecs. So it is possible that if you automatically enable rx immediately after finishing the tx of a mark, that you may start receiving the trailing end of that mark.

However it should be possible to code around this.

See our blog post on this:
http://www.analysir.com/blog/2014/03/27/infrared-receivers-signal-lag/

z3t0:
@AnalysIR In that case should I dismiss this as a PR?

klasse85:
@AnalysIR I would argue that IR receiver lag is something to bare in mind when using the library. My opinion is still that a resume method should make sure that the receiver can receive IR-signals even if the lib has been used to send IR-signals.

AnalysIR:
I just thought I would help, by making the information available. Better, if you guys decide how to incorporate PRs into the library. Just make sure to code & test for the issue I highlighted. Having to re-enable rx after tx is a common 'gotcha' with the library (not just for 'newbies'), but presumably designed-in originally for a good reason. 

It would be great if this approach worked without issue - but I haven't tested it on any or indeed a range of different IR receivers.

z3t0:
Closing as manually re enabling seems like the better approach in this situation

PR #175, opened 2015-07-21
Corrected conditional code to be able to disable some (or all) protocols

jorgediez:
Some minor changes in conditional code to be able to compile the library if some (or all) protocols are disabled in IRremote.h


z3t0:
Please rebase off #107 experimental

PR #183, opened 2015-07-31
Added suport for user defined IR reception feedback LED

madmalkav:
I'm a newbie both in coding and electronics, so I don't know if this is the proper way of doing it.


z3t0:
@madmalkav Thanks, your contribution is greatly appreciated!

PR #193, opened 2015-08-11
Change float operations to int to save program memory

madmalkav:
As suggested in
https://ubiyubix.wordpress.com/2012/05/05/porting-the-arduino-irremote-library-to-the-attiny4313/
there is a series of float operations that can be converted to int
operations reducing final program size


z3t0:
@madmalkav Has this been tested using Receive and Send?

madmalkav:
Only receive. I have no way to test the send. I *think* it shall not affect
but as my previous commit, I'm a newbie and everything I send wiol be
better supervised

z3t0:
@madmalkav Okay no problem, I will test it sometime soon and then merge. Thanks for the contribution

madmalkav:
Can someone please try this? The space saving is quite interesting , specially for the smaller chips.

madmalkav:
Sure, we can use values as big as we want, the thing is avoiding floating point operations for saving space in the final program.

pcoughlin:
Good ... then it would be good to change that one back to INT to keep it from breaking other people's code that currently uses bigger buffers.

madmalkav:
Sorry but as I said I'm not a veteran code, I'm looking at my commit and sorry but I can't figure what isn't being used as an unsigned char instead of an INT on it.

madmalkav:
No len or hz at all in this commit. I saw some comments of you about it in another old commit in my copy of the repository unrelated to this pull request. Sorry if I'm misunderstanding something.

AnalysIR:
there seems to be some changes to sendRAW function parameters in the links above & elsewhere

madmalkav:
Ok, I think I know where this confusion comes from. If you go to the Commits tab it shows some previous activity, the files you are looking at are related to a sync I made from the official repo to mine to be sure my code merged ok. Forget about that and look into the Files Changed tab, that will show you the real changes for this commit.

AnalysIR:
@madmalkav OK so ignore my previous comments.

```  if      (newval < (int) (( (long) ((oldval * 80)/100) ))) return 0 ;```
However, in the above I suspect you may end up with a FLOOR operation vs ROUND

```  if      (newval < (int) (( (long) (((oldval * 80)+50)/100) ))) return 0 ;```
do you think something like this would be an improvement., to get a reasonable rounding of the result?

If so, you could apply to the other statement as well.

madmalkav:
Honestly, I'm not sure, I barely understand the protocols or the maths involved, I just adapted the operations to make in a way I thought the results will be faithful enough and tried ok on my project that previously won't fit my mcu because of the addition of the float point libraries.

Please fell free to change it whatever way you think is more optimal.

adamlhumphreys:
@madmalkav, have you noticed an actual change in program size? (I'd assume as much.) I'd think the compiler would optimize these into fixed variables automatically, much like it would take something like "x / 2" and convert it to "x>>1". I've made such an optimization manually and noticed no change in program size, so you have me curious. I've also tried different comparison operators of '<=' vs. '<' without change.

madmalkav:
I haven't touched that in months, but the difference in final size was
noticeable. If program size isn't reduced perhaps you still have some float
operations somewhere.

adamlhumphreys:
I didn't figure you have based on the dates, but you got me curious since I just made a pull request that restores the macros in place of the MATCH functions and also noticed a fair decrease in program size. So I'll look into this as well and thank you!

ArminJo:
Includes in 2.5.0

PR #198, opened 2015-08-15
update custom_delay function

pcoughlin:
Use micros() to delay based on "real-time" instead of approximation with delay() or delayMicroseconds()

Changed name to custom_delay_usec() to correspond to MicroSeconds ("_ms" is MilliSeconds).

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -253,7 +253,7 @@ class IRsend
 	public:
 		IRsend () { }
 
-		void  custom_delay_ms (unsigned int time);
+		void  custom_delay_usec (unsigned long uSecs);
 		void  enableIROut 		(int khz) ;
 		void  mark        		(int usec) ;
 		void  space       		(int usec) ;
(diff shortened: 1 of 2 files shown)


z3t0:
Thanks for the contribution

AnalysIR:
@pcoughlin 
Just made a PR and one of the changes was a reversal to the "asm" workaround.

Reason: it is not optimized out on my Arduino IDE 1.65.r2, when removed.

It would be good if someone else could verify the need for it.

Alternatively, you could try using the 'volatile' keyword before the function variables, instead - which should stop the optimisation-out occuring. Obviously cannot test this on my system..

Is it possible you have non-standard optimisations in your Arduino(?) IDE?

z3t0:
@AnalysIR Merged your PR. I will take a look at the optimization in question and see if I can learn anything about it

pcoughlin:
@AnalysIR So what I read is that "for you" the line of code makes no difference?

How did you test the two different routines?
Can you post the test Sketch that evaluated the difference so that I can run it on my setup?

Your other changes are on point re: variables and the ranges did need to be larger.
Thanks for your contributions, they are quite good.

AnalysIR:
@pcoughlin Correct & in the meantime another user has confirmed it works OK.

As I mentioned, it is not possible for me to test as the compiler does not optimise-out on my setup.

It works for me with and without the "ASM" workaround.

Did you try using the volatile keyword before the variables.?

PR #201, opened 2015-08-19
Updates to IRremote.h, irSend.cpp

AnalysIR:
Allows, longer marks, spaces by using unsigned int in function parameters
Allows longer signals, reversing a recent change which limited the max signal length to 255, using sendRAW.
Reversed the 'asm' workaround recently made to custom_delay_usec, as removing this workaround does not cause a problem on my Arduino IDE (1.6.5-r2). The original author should check again - maybe there are special compiler optimixzations at play? Alternatively, try using the 'volatile' keyword before the function variables, to prevent the code being optimized out???.

I have tested these changes and verified correct operation on Arduino  IDE (1.6.5-r2) & oscilloscope.

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -255,9 +255,9 @@ class IRsend
 
 		void  custom_delay_usec (unsigned long uSecs);
 		void  enableIROut 		(int khz) ;
-		void  mark        		(int usec) ;
-		void  space       		(int usec) ;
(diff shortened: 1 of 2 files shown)


AnalysIR:
..sorry...some duplication here, only 2 files updated..(not used to github PRs)

z3t0:
@AnalysIR Thank you very much for the pull request. Will merge in a minute

PR #206, opened 2015-08-24
Create IRsendRawDemo

AnalysIR:
As sendRaw is a very popular method of sending unknown or AC IR signals, I thought it would be very useful to include an example of same with the library.

The code in this new example is tested and verified as working with the latest Master.


z3t0:
Thanks

Issue #167, opened 2015-07-03
IRrecvdump - bug & solution


AnalysIR:
HI Guys

There is an annoying set of bugs in this IRrecvdump example provided with the library, which I know causes a lot of hassle & confusion for beginners etc. 

1. The dump output always starts with a space which is generally not required and confusing to many users, particularly as sendRAW must always start with a mark.
2. The line `Serial.print(-(int)results->rawbuf[i]*USECPERTICK, DEC);` for printing out spaces as negative values has a bug in it (It handles large values incorrectly and prints them out as positive)

I have updated the example as follows, which provides a more readable output and fixes the 'bugs' mentioned above. I have also adjusted the formatted output to be more readable for users.
Essentially, just starts outputting at the second value in rawbuf, as the first is always a space & confusing. Also added in a blank line between signals for improved readability.

With this new code the 'dump' will always start and end with a mark, which is preferrable.

Code below is much improved & is tested.
```c
/*
 * IRremote: IRrecvDump - dump details of IR codes with IRrecv
 * An IR detector/demodulator must be connected to the input RECV_PIN.
 * Version 0.1 July, 2009
 * Copyright 2009 Ken Shirriff
 * http://arcfn.com
 * JVC and Panasonic protocol added by Kristian Lauszus (Thanks to zenwheel and other people at the original blog post)
 */

#include <IRremote.h>

int RECV_PIN = 2;

IRrecv irrecv(RECV_PIN);

decode_results results;

void setup()
{
  Serial.begin(9600);
  irrecv.enableIRIn(); // Start the receiver
}

// Dumps out the decode_results structure.
// Call this after IRrecv::decode()
// void * to work around compiler issue
//void dump(void *v) {
//  decode_results *results = (decode_results *)v
void dump(decode_results *results) {
  int count = results->rawlen;
  if (results->decode_type == UNKNOWN) {
    Serial.print("Unknown encoding: ");
  } 
  else if (results->decode_type == NEC) {
    Serial.print("Decoded NEC: ");
  } 
  else if (results->decode_type == SONY) {
    Serial.print("Decoded SONY: ");
  } 
  else if (results->decode_type == RC5) {
    Serial.print("Decoded RC5: ");
  } 
  else if (results->decode_type == RC6) {
    Serial.print("Decoded RC6: ");
  }
  else if (results->decode_type == PANASONIC) {	
    Serial.print("Decoded PANASONIC - Address: ");
[...]

PR #207, opened 2015-08-24
Bug fixes as per Issue #167

AnalysIR:
Have updated IRrecvDdump to fix bugs described in Issue: #167

In summary, removed bug where large space values were displayed incorrectly & confusing users. The output now always starts with a mark, instead of a space, which makes it easier to interpret and less confusing for users.

refer to #167 for more detials.

The update has been tested with several protocols (but not all) and verified as working.

### DIFF
--- a/examples/IRrecvDump/IRrecvDump.ino
+++ b/examples/IRrecvDump/IRrecvDump.ino
@@ -10,7 +10,12 @@
 
 #include <IRremote.h>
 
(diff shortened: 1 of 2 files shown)


AnalysIR:
Changes to IRrecvdumpV2 have been tested with several protocols (but not all) and verified as working.

z3t0:
Thank you for the contribution

PR #212, opened 2015-08-27
LG_AC : heating option

chaeplin:
1) heating : using rawcode at https://github.com/chaeplin/RaspAC/blob/master/lircd.conf
2) description to decode LG AC : IRrecvDumpV2 and AnalysIR's code
3) remote of LG AC has two type of HDR mark/space, 8000/4000 and 3100/10000


z3t0:
Thanks for the contribution!

PR #213, opened 2015-08-28
Update IRrecvDumpV2.ino

AnalysIR:
- Cleaned up layout of output, removed trailing commas
- buffer sizes are now correct
- no need to add 0 for space at end, as sendRAW takes care of this


z3t0:
Thanks

Issue #214, opened 2015-08-31
Issues when compiling IRremote with Arduino-Makefile


audetto:
Hi,

I've been using for a while the Makefile for Arduino 

https://github.com/sudar/Arduino-Makefile

but recently had an issue where IRremote would be miscompiled.
Compiles, links, uploads, but hangs in

irrecv.enableIRIn();

everything started here

https://github.com/z3t0/Arduino-IRremote/commit/88e243fe068e06d2ed602fefe6bd5effd07006e1

when the library was split in many files.

(issue for Arduino-Makefile: https://github.com/sudar/Arduino-Makefile/issues/376)

Do you know guys of anything that would cause issues in the new layout of IRremote code?
Comparing the Makefile with Arduino IDE 1.6.5, I can see that the output of the Makefile is a bit smaller (0.5KB), while before the refactoring they were the same.

Probably some obscure compiler or linker flag. Just checking if it rings a bell.

audetto:
I've found what the issue is (long description in the linked issue).
Basically irISR.cpp exports no symbols so accordingly to the tool chain setup it might get included or not.

In the Makefile, a static lib of the libraries used is created, so at the end, the linker does not have any reason to include it in the final output.
The IDE does link in a different way and it works.

Probably a weak feature to rely on, and forcing the linker to include the ISR is not difficult.

I would propose to merge irISR.cpp back into irRecv.cpp so whenever the receive code is linked in, the ISR is set no matter what the linker chooses.

(I seem to understand the ISR is only needed to decode, otherwise the best location would be some file that gets included all the times)

I will prepare a pull request: what do you think?

z3t0:
@audetto Sure, please prepare a PR and I will merge it. I was not aware of this problem because I use the Arduino IDE to compile and upload, thanks for pointing it out.

audetto:
I've prepare the PR.
Is it good? Do you like it?

audetto:
Perfect. All done, working now.

z3t0:
No problem, thanks for the contribution

PR #215, opened 2015-09-02
Merge back irISR.cpp into irRemote.cpp to avoid a linker issue

audetto:
https://github.com/z3t0/Arduino-IRremote/issues/214

I thought to put the code in irRemote.cpp as this is where the global variable used by the ISR is: irparams.


z3t0:
Does this work for you? Can you please share the output of the makefile? Thanks

audetto:
100%: this works perfectly for me now.

Here is the makefile:

==================================================================
BOARD_TAG=uno
ARDUINO_DIR=/home/andrea/projects/pi/ide/arduino-1.6.5
ARDUINO_SKETCHBOOK=/home/andrea/Arduino

include /home/andrea/projects/cvs/3rdParty/Arduino-Makefile/Arduino.mk
==================================================================

and output at http://pastebin.com/yhshWRwR

Now, irISR.cpp (which does not exist ay longer) is always linked in the final executable (with the TIMER code) as it is part of irRemote.cpp which is required to satisfy linker dependencies (the linker cannot split a .o, it can only decide to include it or not).

Builds with Arduino-IDE or Arduino-Makefile have both same size (+/- 10 bytes) like before the split in IRremote (same size one to the other, not before after).

z3t0:
Thanks

PR #219, opened 2015-09-09
Changed library name "RobotIRremote"

bessl:


pcoughlin:
Good catch.  Just a case change to match the library name.

Please add a description for the pull request.

z3t0:
thanks for the correction!

PR #227, opened 2015-10-25
Change definition of IRsend::sendRaw array parameter to const

Sebazzz:
Change definition of IRsend::sendRaw method array parameter to const to allow passing constant array to function. This would allow passing const arrays to the method.


z3t0:
@Sebazzz but would that not restrict you to only passing constant arrays?

Sebazzz:
No but it would allow me to pass constant arrays. `const` in this context only means that the function won’t alter the array, which it doesn’t.

PR #228, opened 2015-10-29
Add Seiki TV Send Support

mattbucci:
Add Seiki Support based on lirc.conf. Confirmed working

Predata based on AIWA, so thanks for that.

```
# Please make this file available to others
# by sending it to <lirc@bartelmus.de>
#
# this config file was automatically generated
                                                                                                                                                                                 24,1           2%

# Please make this file available to others
# by sending it to <lirc@bartelmus.de>
#
# this config file was automatically generated
# using lirc-0.9.0-pre1(default) on Tue Oct 20 08:30:27 2015
#
# contributed by
#
# brand:                       /root/lircd.conf
# model no. of remote control:
# devices being controlled by this remote:
#

begin remote

  name   seiki
  bits           16
  flags SPACE_ENC|CONST_LENGTH
  eps            30
  aeps          100

  header       9057  4467
  one           606  1646
  zero          606   520
  ptrail        607
  pre_data_bits   16
  pre_data       0x40BE
  gap          108167
  toggle_bit_mask 0x0

      begin codes
          KEY_POWER                0x629D
          KEY_VOLUMEUP             0x30CF
          KEY_VOLUMEDOWN           0x9867
          KEY_CHANNELUP            0xF00F
          KEY_CHANNELDOWN          0x5AA5
          KEY_SWITCHVIDEOMODE      0xD22D
          KEY_ENTER                0x52AD
      end codes

end remote
```

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -76,6 +76,9 @@
 #define DECODE_PRONTO        0 // This function doe not logically make sense
 #define SEND_PRONTO          1
 
+#define DECODE_SEIKI        0 // NOT WRITTEN
+#define SEND_SEIKI          1
+
 //------------------------------------------------------------------------------
 // When sending a Pronto code we request to send either the "once" code
 //                                                   or the "repeat" code
@@ -115,6 +118,7 @@ typedef
 		SHARP,
 		DENON,
 		PRONTO,
+		SEIKI,
 	}
 decode_type_t;
 
@@ -242,6 +246,10 @@ class IRrecv
 #		if DECODE_DENON
 			bool  decodeDenon (decode_results *results) ;
 #		endif
+		//......................................................................
+#		if DECODE_SEIKI
+			bool  decodeSeiki (decode_results *results) ;
+#		endif
 } ;
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


bengtmartensson:
Loading this file in IrScrutinizer shows that it is not a newly discovered protocol, but the NEC protocol ("NEC2" to be precise, the whole signal is repeating); device number 2, subdevice 125, function codes 70,12,25,15,90,75,74 respectively. (Note that Shirriff reads the parameters in the other bit order: 2 = 00000010 -> 01000000 = 64 etc.) 

So there is no need to add it.

mattbucci:
@bengtmartensson That seems rather confusing considering that this is a 16bit pattern and NEC is 32bit, but I don't know enough about it. 

I just got it working with my TV and figured I'd share. Especially considering how the predata threw me off for an hour or so.

bengtmartensson:
The "missing" 16 bits are contained in Lirc's pre_data. Since they are the same for all the signals contained therein, Lirc bundled them into pre_data.

Let me suggest that you try [IrScrutinizer](http://www.harctoolbox.org/IrScrutinizer) instead of Lirc for identifying your signals.

PR #253, opened 2016-01-07
Make sendPronto works with IRsend

ValeraVi:


z3t0:
I'll wait on confirmation from other users

irPronto.cpp
```diff
@@ -62,20 +62,24 @@ uint8_t  htob (char ch)
...
 }
 
 //+=============================================================================
 //
-bool sendPronto (char* s,  bool repeat,  bool fallback)
+bool
+#if !TEST
+	IRsend::
+#endif
+	sendPronto (const char* s,  bool repeat,  bool fallback)
```
cymerio: We need to make `sendPronto` a method of `IRsend`. Otherwise it will not be accesible from outside this file.
If we are in `TEST` mode, `sendPronto` is a standalone function and `irPronto.cpp` is a standalone program.

z3t0:
Thank you for the confirmation. However, I think it might be a good idea to put this PR on hold considering that we are thinking about rewriting the way Pronto is implemented in this librayr #502

ArminJo:
Changes are included, but pronto is currently disabled.

PR #258, opened 2016-01-24
Update IRremote.cpp to improve debugging

ElectricRCAircraftGuy:
This is a small change, and definitely an improvement. I simply improved the debugging by stating whether a check passed or failed, for easier identification in debug mode.

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -50,9 +50,9 @@ int  MATCH (int measured,  int desired)
 
   bool passed = ((measured >= TICKS_LOW(desired)) && (measured <= TICKS_HIGH(desired)));
   if (passed)
-    DBG_PRINTLN(F("; passed"));
+    DBG_PRINTLN(F("?; passed"));
   else
-    DBG_PRINTLN(F("; FAILED")); 
+    DBG_PRINTLN(F("?; FAILED")); 
  	return passed;
 }
 
@@ -65,7 +65,7 @@ int  MATCH_MARK (int measured_ticks,  int desired_us)
 	DBG_PRINT(measured_ticks * USECPERTICK, DEC);
 	DBG_PRINT(F("us vs "));
 	DBG_PRINT(desired_us, DEC);
-  DBG_PRINT("us"); 
+	DBG_PRINT("us"); 
 	DBG_PRINT(": ");
 	DBG_PRINT(TICKS_LOW(desired_us + MARK_EXCESS) * USECPERTICK, DEC);
 	DBG_PRINT(F(" <= "));
@@ -76,9 +76,9 @@ int  MATCH_MARK (int measured_ticks,  int desired_us)
   bool passed = ((measured_ticks >= TICKS_LOW (desired_us + MARK_EXCESS))
                 && (measured_ticks <= TICKS_HIGH(desired_us + MARK_EXCESS)));
   if (passed)
-    DBG_PRINTLN(F("; passed"));
+    DBG_PRINTLN(F("?; passed"));
   else
-    DBG_PRINTLN(F("; FAILED")); 
+    DBG_PRINTLN(F("?; FAILED")); 
  	return passed;
 }
 
@@ -91,7 +91,7 @@ int  MATCH_SPACE (int measured_ticks,  int desired_us)
 	DBG_PRINT(measured_ticks * USECPERTICK, DEC);
 	DBG_PRINT(F("us vs "));
 	DBG_PRINT(desired_us, DEC);
-  DBG_PRINT("us"); 
+	DBG_PRINT("us"); 
 	DBG_PRINT(": ");
 	DBG_PRINT(TICKS_LOW(desired_us - MARK_EXCESS) * USECPERTICK, DEC);
 	DBG_PRINT(F(" <= "));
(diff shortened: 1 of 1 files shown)


IRremote.cpp
```diff
@@ -41,14 +41,19 @@
 //
 int  MATCH (int measured,  int desired)
 {
- 	DBG_PRINT("Testing: ");
+ 	DBG_PRINT(F("Testing: "));
```
z3t0: What does F("") do? I'm not a very advanced C programmer so can you please explain it to me? Thanks!

ElectricRCAircraftGuy: It's an Arduino PROGMEM macro, to cause the string constant to take up Flash memory only but not SRAM. See here at the bottom: http://playground.arduino.cc/Learning/Memory

ElectricRCAircraftGuy: Without it, all printed string constants take both flash and RAM memory

ElectricRCAircraftGuy: For anything <2 chars it's a wash. Just using it takes a couple bytes.

z3t0: Thank you for the detailed explanation. I will look into merging this soon. Once again thank you for the contribution

IRremote.cpp
```diff
@@ -41,14 +41,19 @@
...
  	DBG_PRINT(TICKS_LOW(desired), DEC);
- 	DBG_PRINT(" <= ");
+ 	DBG_PRINT(F(" <= "));
  	DBG_PRINT(measured, DEC);
- 	DBG_PRINT(" <= ");
- 	DBG_PRINTLN(TICKS_HIGH(desired), DEC);
+ 	DBG_PRINT(F(" <= "));
```
pcoughlin: When using ```F()``` does the compiler optimize the exact same strings and only put them into the Flash once?
IF not, then we should look at a way to make that happen (not just here, but review throughout).

Also, using ```F()``` For all the text is a great idea.  The whole Library should probably have that change applied. Kudos to @ElectricRCAircraftGuy

IRremote.cpp
```diff
@@ -56,19 +61,25 @@ int  MATCH (int measured,  int desired)
...
-	DBG_PRINT(TICKS_LOW(desired_us + MARK_EXCESS), DEC);
-	DBG_PRINT(" <= ");
-	DBG_PRINT(measured_ticks, DEC);
-	DBG_PRINT(" <= ");
-	DBG_PRINTLN(TICKS_HIGH(desired_us + MARK_EXCESS), DEC);
+	DBG_PRINT(TICKS_LOW(desired_us + MARK_EXCESS) * USECPERTICK, DEC);
+	DBG_PRINT(F(" <= "));
+	DBG_PRINT(measured_ticks * USECPERTICK, DEC);
+	DBG_PRINT(F(" <= "));
+	DBG_PRINT(TICKS_HIGH(desired_us + MARK_EXCESS) * USECPERTICK, DEC);
```
pcoughlin: The addition of ```* USECPERTICK``` will make the output numbers TIME instead of TICKS and displayed bigger numbers..  
For instance (USECPERTICK=50):  12 -> 600 or 210 -> 10500

There is an argument for consistency. Many other outputs are TIME, not TICKS.

What do you think?

ElectricRCAircraftGuy: I think, to a user looking for bugs, time is more readable, especially since what I'm comparing to is time, based on the standards for each protocol. Ex: [NEC standard is 562.5us](https://sibotic.files.wordpress.com/2013/12/adoh-necinfraredtransmissionprotocol-281113-1713-47344.pdf) (IRremote uses 560), which is what I'm looking for. Ticks requires me to do a calculation first--it's not as intuitive for debugging, I think.

IRremote.cpp
ElectricRCAircraftGuy: Q: "When using F() does the compiler optimize the exact same strings and only put them into the Flash once?"
A: No, as far as I can tell, even if the strings are identical, they are put in the flash again each time you use the F() macro. It's not an ideal solution, just a better solution than not using F(). To put it in only once you'd have to manually do that, then manually call them out to print them using AVRLibc directlyhttp://www.nongnu.org/avr-libc/user-manual/group__avr__pgmspace.html.

pcoughlin: Since the strings are used several times each, why not put them in like this:

```
const char DBGtext_lteq[] PROGMEM  = {" <= "};
const char DBGtext_pass[] PROGMEM  = {"?; passed"};
const char DBGtext_fail[] PROGMEM  = {"?; FAILED"};
```
Then output them like this:
``` DBG_PRINTLN(F(DBGtext_pass)); ```

I am not sure if the F() should be used in the above or not.

Would this solve the duplicate string issue and gain that part of the PROGMEM back as well?

IRremote.cpp
pcoughlin: OK ... I will go along with that.

z3t0: Yes this does seem like a good idea

IRremote.cpp
z3t0: #261

reednoel4u:
Some body help me add the library, I follow the instruction but can't found the folder with name IRremote, only two file IRremote.h and IRremote.cpp 
"4. Move the "IRremote" folder that has been extracted to your libraries directory."
Sorry my english and I'm a newbie.
Rgs

z3t0:
@reednoel4u If you need help please create a new issue. Thanks!

PR #259, opened 2016-01-25
Added TCL112AC A/C unit. 112 bits and bigger BUFFER

caspercba:
This code works fine, but of course I have the following issues and I wouldn't like to merge until we define how to solve them.

Im creating this pull request just to show the reviewer the problems im facing:
1. Panasonic decode algorithm does not check the data length, and because this protocol is very similar (112 bits instead of max 48bits in panasonic), sometimes its cached as panasonic and truncated to 48bits. Should we add a size check in panasonic?
2. Raw buffer length is not enough to hold rawData[227]. Of course i don't want to change this for every user. What would you suggest?

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -28,6 +28,10 @@
 // Each protocol you include costs memory and, during decode, costs time
 // Disable (set to 0) all the protocols you do not need/want!
 //
+
+#define DECODE_TCL112AC      1
+#define SEND_TCL112AC        1
+
 #define DECODE_RC5           1
 #define SEND_RC5             1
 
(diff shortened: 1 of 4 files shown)


pcoughlin:
I do not see the length check code for the Panasonic decode that you recommend in your comments above.

Am I missing it?  If so, then please give file/line I could review.  Thanks.

caspercba:
Hi,
issue #1 is a question. I guess the answer is yes, we want size check in panasonic code. Ill create a different Pull Request for this.
Cheers.

irfn:
@caspercba can you reference the pull request here once you are done.

PR #268, opened 2016-02-10
Proposed 2.1.1 Update

adamlhumphreys:
Greetings @z3t0,
It sure took me long enough, eh?

I think I have the macros fully restored now. Initially, I had placed them back in IRremoteInt.h, but when I tried to compile in DEBUG mode, it complained about the lines where the function prototypes are located. Removing the macros in IRremoteInt.h enabled it to compile, so then I placed the macros in the IRremoteInt.cpp under the DEBUG functions, but then trying to compile without debug, it complained when it saw MATCH et al. not being defined in ir_Aiwa.cpp, the first file in alphabetical order, because it wasn’t defined until IRremoteInt.cpp. Moving the macros right under the DEBUG function prototypes in IrremoteInt.h seems to do the trick for both compile modes! :D

For reference I’m currently using Arduino 1.0.1 with WinAVR-20100110 with a Teensy 2.0, (will be using decodes on Arduino NANOs though). Here’s my edits in a nutshell:
### Changes
- Updated IRrecvDumpV2.ino to include gap reading to determine send repeat timing
- Fixed false positive REPEAT in Sony and Sanyo protocols due to code errors
- Restored macros for MATCH, MATCH_MARK, and MATCH_SPACE for efficiency
- Updated protocols with separate “offset++;” in place of “[offset++]” to function with macros
- Optimized by reducing floating point operations as suggested by @madmalkav (#193)
- Optimized with macros when using default USECPERTICK and TOLERANCE
- Made decodeHash as a settable protocol defined by DECODE_HASH
- Fixed some minor typos
### Additions
- Added Jensen Protocol [@adamlhumphreys]
- Added Heater Protocol [@adamlhumphreys]
- Added “SHUZU_” prefix to macro variables in ir_Template.cpp to maintain distinction and standardization
- Added notice to use “offset++;” in place of “[offset++]” in ir_Template.cpp for macros
- Added missing semicolon found by @mattman00000 (#271)


z3t0:
Awesome, looks like you were busy! I will get to it as soon as possible.

Thanks again for the contribution, it really does make a difference!

Thanks,
Rafi Khan


On Wed, Feb 10, 2016 at 2:41 AM, Adam L. Humphreys <notifications@github.com

adamlhumphreys:
I did what I could. :) I'll also be taking a look at @madmalkav's optimizations and probably incorporate them into my current branch when I get the chance, (but I could do that after a pull too, either way). I think I can get rid of the additional division as well. Much appreciated!

No probs! I've been having fun with this library. I have a feeling I might just get lazy when it comes to using separate hardware buttons in prototyping!

-ALH

z3t0:
note to self;
TODO: add #271 to contributors list

z3t0:
@adamlhumphreys Would it be possible for you to rebase this on the master as I am keen on merging this in? Otherwise I can do it just that it may take a while for me to get to it...

adamlhumphreys:
If you mean update my update with the current master 2.1.0 version, I did that Feb. 21, (locally as well), hence the new pull request name of 2.1.1. But I did notice you changed the library.json file, so I just now updated that. (Had to figure out GitHub again, at least I was much quicker this time. ^^) So it should all be good to go.

I've also developed a Fast Decode based on binary masks that can, (with less rigorous checking and tweak-ability), take the place of NEC, Sony, Jensen, and whatever else uses a typical IR protocol, (even some previously unrecognized codes from an old universal remote we had programmed for something). But if the marks aren't present as expected, it returns false. I plan to experiment with it a little more, and maybe add some better optional header recognition, (which is tricky without distinct headers or variable lengths). But for my purposes I just ignore the header and get a consistent code rather than a hash. :)

adamlhumphreys:
@z3t0 Everything should be good to go. :)

z3t0:
Sorry for the delay. Thanks again for all the hard work

z3t0:
If possible please rebase it so I can merge

adamlhumphreys:
That's been done, (a couple of months ago actually). Does it need to be done again or is everything good?

z3t0:
It seems like theres some sort of conflict as I can't merge it directly. I will take a look in a few days and see if its anything problematic.

adamlhumphreys:
Huh. The only conflict it tells me on this thread is "Only those with write access to this repository can merge pull requests," of course that goes without saying. I guess GitHub isn't quite as streamlined as we'd like for it to be. ^^ Thank you for your efforts! :)

PR #276, opened 2016-02-21
sync

z3t0:


z3t0:
things have gotten rather messy, is there anything in this branch that needs to be preserved? otherwise it would be nice to wipe the slate clean

PR #288, opened 2016-02-23
Add sendRaw with byte[] and IR code transcoder Demo

shadwork:
Add sendRaw and sendRawDelayed with byte array parameters to avoid memory consume by int.
sendRawDelayed add delay after sending start, so you can receive ir right after transmission without problem.
Two new examples:
IRrecvDumpRawByte - Dump raw data in byte form for sendRaw and sendRawDelayed with byte array
DaewooR40A01toDVDHR755Transcoder - Receive ir from DAEWOO R-40A01 and transmit to Samsung DVD-HR755


shadwork:
sorry. forget add new examples to .travis.yml

brunoudi:
I saw on IrScrutinizer a format "UEI Learned", and it looks like byte[].

Is there a way of use this type, or convert this "UEI Learned" to a byte[] for use like this issue?

z3t0:
@bengtmartensson

bengtmartensson:
@brunoudi;
It is completely unclear to me what problem you are addressing, or what you are trying to achieve. (Please explain.) "UEI learned" is the format UEI remotes internally use for storing its learned signals, see [the JP1 forum](www.hifi-remote.com/forums/), but it is almost surely not the solution to your problem.

brunoudi:
@bengtmartensson Sorry About the unclear question..

The point is I'm trying to build a Arduino to control a lot of devices in my house. 

Considering the quantity of differents commands, my idea is to build a "generic sketch" where it receives the raw data (or other format) by a UDP packet and sends the IR Command.

After studying IR coding, I realize the Raw Data can be quite big, and because of that it is not good to put inside a UDP packet.

This alternative, to use byte[] to avoid memory consume (basically divides de data by 50) sounds goods, but after get IR Codes at http://irdb.globalcache.com/ and convert it using IrScrutinizer, I get a Raw Data where not all values are multiple of 50.

So, It is not clear to me a best way of doing this. The IR command expressed in UEI Learned format looks quite simple, but I believe I will not be able to convert it to a Raw Data and use with IR Remote.

bengtmartensson:
Ok that is easy. Adding new export formats to IrScrutinizer is a breeze... 

* Make sure you are using IrScrutinizer 1.3 or later
* Drop [C-raw-ticks.xml](https://gist.github.com/bengtmartensson/36f8abe3a2951e3f45eb9cc8f567271f) into your exportformats.d directory. (This is just a simple variation of the C export format.)
* Add the lines 

    `<xi:include href="exportformats.d/C-raw-ticks.xml">`
        `    <xi:fallback/>`
    `</xi:include>`
to exportformats.xml.

This will give you a new export format "C raw ticks" that should do what you want. Feel free to modify C-raw-ticks.xml if desired.

brunoudi:
Ok, I got it.. The output of a signal is:

```const int frequency = 38400;
const char *protocol = "necx2";
const int intro_burst_length = 0;
const int repeat_burst_length = 34;
const int ending_burst_length = 0;
const int tick = 50;

const int raw_necx2_7_7_2[] = { 90,90,11,34,11,34,11,34,11,11,11,11,11,11,11,11,11,11,11,34,11,34,11,34,11,11,11,11,11,11,11,11,11,11,11,11,11,34,11,11,11,11,11,11,11,11,11,11,11,11,11,34,11,11,11,34,11,34,11,34,11,34,11,34,11,34,11,930 };
```

The original Raw was:

```
+4512 -4512 +564 -1692 +564 -1692 +564 -1692 +564 -564 +564 -564 +564 -564 +564 -564 +564 -564 +564 -1692 +564 -1692 +564 -1692 +564 -564 +564 -564 +564 -564 +564 -564 +564 -564 +564 -564 +564 -1692 +564 -564 +564 -564 +564 -564 +564 -564 +564 -564 +564 -564 +564 -1692 +564 -564 +564 -1692 +564 -1692 +564 -1692 +564 -1692 +564 -1692 +564 -1692 +564 -46524
```

To make it clear:

1) I realize with raw ticks, the data is rounded... Maybe to up, maybe to down... This rounding does not risk the command? I know the IR Command has tolerance, but can I assume it will be always OK?

2) When I transmit the raw ticks, maybe a value takes 2 bytes (in the example the last value is '930'). Will I need to handle that, or can I just pass the byte[] directly to this function?

bengtmartensson:
So, specifications fulfilled!

To your questions:
1. For all practical and presently existing protocol, 50ms resolution should be enough. That is the resolution of IrRemote as well as Infrared4Arduino and other. This is not to say that 50microseconds is enough for all present and future applications...
2. 255*50microseconds =12750microseconds = 12.75 ms, which "hardly ever" occurs *inside* an Ir sequence. Note that 930 is the final silence. Many systems, including Lirc and IrRemote, just ignore the final gap. However, you have to take care if repeating the signal, for example your NEC2 signal (which repeats every 114ms).

PR #303, opened 2016-03-26
 Fixed bug in ir_Dish.cpp for sending

AnalysIR:
One of our users of AnalysIR, reported issues with sending DIsh signals. After some investigation we realised that this file was neglecting to send the trailing mark after the bits. Fix is included in this update & has been tested on a live Dish device by our own user.

AnalysIR - 26th March 2016
----------------------------------------
https://www.AnalysIR.com/


z3t0:
note: need to add changelog info ... #304

PR #308, opened 2016-04-21
distinguish SANGUINO from MICRODUINO Core+

dmsherazi:
SANGUINO and MICRODUINO both use Atmega_644PA although they have completely different pin mappings. 

PD6 is  PIN 14  on sanguino, while on Microduino it is PIN 8

### DIFF
--- a/IRremoteInt.h
+++ b/IRremoteInt.h
@@ -184,10 +184,10 @@ EXTERN  volatile irparams_t  irparams;
 	#define IR_USE_TIMER2     // tx = pin 1
 	//#define IR_USE_TIMER3   // tx = pin 16
 
-// Sanguino
+// Sanguino  or Microduino
 #elif defined(__AVR_ATmega644P__) || defined(__AVR_ATmega644__)
 	//#define IR_USE_TIMER1   // tx = pin 13
-	#define IR_USE_TIMER2     // tx = pin 14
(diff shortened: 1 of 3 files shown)


z3t0:
@dmsherazi Do you know if other #defines exist for the differing boards?

dmsherazi:
@z3t0  I am not sure about Sanguino but I searched them for MIcroduino and found they haven't provided  any #define for the board.

danny-source:
microduino same uno or Leonardo

Dost Muhammad Shah <notifications@github.com> 於 2016年4月23日 星期六寫道：

dmsherazi:
Microduino Core+ defines NUM_DIGITAL_PINS as 32, whereas Sanguino defines it as 24.
#if defined(AVR_ATmega644PA) && (NUM_DIGITAL_PINS == 32)

MCUdude:
Whoa, this is not the way if down' it. The ATmega644 and the rest of its family all got 32 IO pins, have a look at the [MightyCore](https://github.com/MCUdude/MightyCore). The Sanguino core are not updated anymore, and shouldn't be used. If you're going to add some extra #ifdefs for the Microduino core, it should look something like this:

``` c++
#if defined(MICRODUINO)
  // Do something here
#else
  //Do something else here
#endif

```

z3t0:
Ha! Thanks for the catch. We will wait until a suitable approach is found, is MICRODUINO defined ? in the core?

PR #309, opened 2016-04-27
Add Lego Power Functions send protocol

henkel:


AnalysIR:
You might want to add in a description of which particular Lego device(s) this supports, as I believe there are at least 2 Lego IR protocols?

(Hopefully I didn't miss it?)

henkel:
I implemented the LEGO Power Functions RC protocol. Version 1.10 and 1.20 are identical on the low level implementation that is covered here. Command data could differ. I'm not aware of another Lego Power Functions IR protocol. Do you have a link available?

AnalysIR:
I was thinking of the lego Robotics invention System from 2001 (I still have the IR module).

I didn't realise 'Power Functions' is part of the brand name of this Lego range & assumed it referred to a power signal for Lego. (e.g. [https://en.wikipedia.org/wiki/Lego_Mindstorms](url) )

Sorry for the confusion....and thanks for adding the receiver model #.

henkel:
I updated the changelog to simplify merging

z3t0:
Thank you for the contribution

PR #315, opened 2016-05-24
Send Codes by direct wiring (without IR-LED)

vitalogy:
With these changes in this PR, I can send codes from my Arduino Nano to my Raspberry running lirc and the lirc-rpi overlay, without to use a IR-LED (Arduino) and IR-receiver (Raspberry).
When `SEND_BY_DIRECT_WIRING` is set to `1` in the header IRremote.h, the `TIMER_PWM_PIN` will be used to send digital codes and not pwm.

Please have look, it is working for me

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -95,6 +95,15 @@
 #define PRONTO_NOFALLBACK  false
 
 //------------------------------------------------------------------------------
+// When set to 1, you can wiring directly to your target device, without
+// to use an IR-LED and IR-receiver.
+// All codes will be send without a carrier and digital.
+// NOTE: If you have a 5V Arduino board you should use a resitor divider to do
+// not damage your target device input pin, you have directly wired.
+//
+#define SEND_BY_DIRECT_WIRING	0
+
+//------------------------------------------------------------------------------
 // An enumerated list of all supported formats
 // You do NOT need to remove entries from this list when disabling protocols!
 //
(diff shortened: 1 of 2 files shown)


vitalogy:
I should mention, I could only test RC5.

z3t0:
This looks really cool! I'll take a look and try setting it up on my side to make sure everythings working fine. Thanks for the unique contribution!

z3t0:
I think there may be a cleaner way to implement this, I'll give it a shot once I get a chance, but what I was thinking of was to first, use something other than digitalWrite (maybe?) and also to do something similar to TIMER_DISABLE_PWM for the wire

bengtmartensson:
My 2 cents:

* The terminology is very badly chosen. I would say that what you want to do is to "generate signals without modulation".
* The main reason for doing this (IMHO) is either some IR protocols not using modulation (Revox, Barco, Archer), or, more interesting, generating RF signals (e.g. 433 MHZ (Europe) using cheap sending  modules (Ebay etc)) for control of power switches etc. I have been doing successfully this for quite some time now.
* In [Infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino) (which is my "competitor" to IRremote :-) ) there are different classes for generating modulated and unmodulated signals.

KristianKorsgaard:
I have made a more comprehensive implementation of this addition in #561

bengtmartensson:
Current master contains a CCP symbol USE_NO_CARRIER that appears to do exactly what the O.P. asks for.

If that is the case, thus PR can be closed.

PR #321, opened 2016-06-22
add disableIRIn()  function

bill0:
I use this lib ，think it is very good, thanks authors!! and I find timer2 is not disable, let other function can not RUN well


bill0:
I use this lib ，think it is very good, thanks authors!! and I find timer2 is not disable, let other function can not RUN well

z3t0:
Please check the travis build as it has failed, I'll take a look into it as I get a chance in the following days.

IRremote.h
```diff
@@ -177,6 +177,7 @@ class IRrecv
 		void  blink13    (int blinkflag) ;
 		int   decode     (decode_results *results) ;
 		void  enableIRIn ( ) ;
+		void  disableIRIn ( )；
```
mariusoei: Pretty sure the problem has to do with the weird-looking semicolon at the end of this line. 
Would love to change this but I don't know how to do that on GitHub...

Issue #319, opened 2016-06-19
supports atmega1284p


marciok6:
Supports atmega1284 ?

MCUdude:
have a look at #324  😉

z3t0:
de3d723574bf0f0f2377fb8fba2c85eb483c3745

PR #323, opened 2016-06-23
Added ATmega1284(P) support

MadGarra:
Added support for ATmega1284 based boards (I.e. Sleeping Beauty / Mighty
Stick)
Based on existing ATmega644. Using Timer2 and PIN14 as send pin, but pin
13 (PD5) instead of pin8 (PD0) as BLINKLED
Tested on Mighty Stick with ATmega1284(Non P ver.)/16Mhz


MadGarra:
Sorry. This Pull Request should be directed to my fork instead of this repository. In any case I'll keep it here, as there is a related open issue #319.

PR #337, opened 2016-07-15
Add ATmega48 and ATmega88 to the list (+ some other minor commits)

MCUdude:
I've just created an Arduino core for the ATmega8, ATmega48, ATmega88, ATmega168 and ATmega328, But the ATmega48/88 are missing on the list. How about adding these? 😃


z3t0:
Thanks for the contribution! Very much appreciated.

PR #343, opened 2016-08-01
Minor fixes for Teensy 3.x

PaulStoffregen:
This fixes compiler errors & warning when building for Teensy 3.x, and updates to more flexible settings to support any clock speed.

Only the Teensy-specific defines in IRremoteInt.h are changed, so no other boards should be impacted by this pull request.


z3t0:
Thank you very much!!!

Issue #333, opened 2016-07-11
Protocols Implementation


z3t0:
@adamlhumphreys @Gustavomurta 

It has come to the attention of many users that many of the protocol implementations are questionable and possibly incorrect. This issue will be used to document the effort to find "reputable" sources for the specifications of the different IR protocols so that they can then be referenced and so the library has a way to explain what arbitrary names such as "LG" really mean.
#331

bengtmartensson:
I am happy that this issue has been brought up. Actually, it is an recurrent issue at remotecentral.com and hifi-remote.com that someone reports e.g. an NEC signal captured by IRremote, but the parametrization is different from the rest of the world, creating a fair amount of misunderstanding. 

I have been working on maintaining a list of IR protocols, [this one](https://raw.githubusercontent.com/bengtmartensson/harctoolboxbundle/master/IrpMaster/src/main/config/IrpProtocols.ini). It is in machine readable form, in the so-called [IRP notation](http://www.hifi-remote.com/wiki/index.php?title=IRP_Notation) a slightly cryptical, man- and machien readable notation for IR protocols. It is used in IrScrutinizer, which can generate the raw IR signals from the IRP form in the data base. It is based on the library DecodeIR by John Fine, with subsequent improvements by the [JP1 community](http://www.hifi-remote.com/forums). So I personally do not take credit for it, nor claim copyright. It is therefore compatible with DecodeIR (which is a library for IR signal decoding, used in IrScrutinizer, RemoteMaster, IrScope), and the "general consensus" on IR protocols and their interpretations in the JP1 community. It does not represent the "truth", for example in the sense of a standard by a standardization organization like ISO. IR protocols are not registered by a central registration, so there is a fair amount of arbitrariness involved. (For example, the protocol therein referred to as "denon" is by the company Denon called "Sharp" (!)). 

This is not a finalized list; improvements are welcome (possibly in particular to the documentation), but it should be kept consistent with the JP1 community and DecodeIR.

So if you are considering "canonicalizing", please consider being compatible with this list. Work is in progress for generating decoder/rendering code directly from this list... 

"LG" corresponds to "Nec2" in "my" list.

z3t0:
@bengtmartensson 

Thanks for the detailed comment, I will look into IRP Notation and see if I can find a way to make it work with this library. Also I think it would be a great idea to base it off some sort of community standard specification (such as "your"  list) for the community is the main target in either case.

[...]

PR #347, opened 2016-08-11
incorrect times

Wietse012:
Mark & space time are incorrect according the following site
http://www.sbprojects.com/knowledge/ir/jvc.php
(tested on my JVC amp, works fine with the edited times)

### DIFF
--- a/ir_JVC.cpp
+++ b/ir_JVC.cpp
@@ -10,8 +10,8 @@
 //==============================================================================
 
 #define JVC_BITS           16
-#define JVC_HDR_MARK     8000
-#define JVC_HDR_SPACE    4000
+#define JVC_HDR_MARK     8400
+#define JVC_HDR_SPACE    4200
 #define JVC_BIT_MARK      600
 #define JVC_ONE_SPACE    1600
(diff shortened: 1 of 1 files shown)


z3t0:
#333

AnalysIR:
FYI: the mark/space times for 0/1 data  bits also seem to differ.

Based on my info - the timings from sbprojects (manually measured on a scope) is more accurate than currently defined in IRremote

Wietse012:
Thats true, i't didn't notice me directly but now i changed the bit_mark, one_space and zero_space also, my raw data (from IRtransmitter) on IRreciever looks way more like the remote control signal.

#define JVC_BITS                  16         //old values
#define JVC_HDR_MARK       8400   //8000
#define JVC_HDR_SPACE      4200   //4000
#define JVC_BIT_MARK         526      //600
#define JVC_ONE_SPACE      2100    //1600
#define JVC_ZERO_SPACE    526      //550
#define JVC_RPT_LENGTH    60000

bengtmartensson:
Suggest closing, surpassed by #333.

PR #351, opened 2016-08-22
Move ATmega32U4 to timer3 and pin 5 output for compatibility with the…

slacker87:
… arduino pro micro


z3t0:
I think it would be better to see if there is a define passed to specifically detect the arduino pro micro, as chasing the pin would then make the comment useless for Teensy users?

PR #352, opened 2016-08-23
Move board specific configuration info to new file boarddefs.h.

bengtmartensson:
This PR proposes to move board definitions to its own file, `boarddefs.h`. There are two reasons for this change:
1. Code organization and maintenance: Presently, a lot of different stuff is lumped into `IRremoteInt.h`: typedefs, defines of different types including user preferences, and a "data base" of board specific information. This is not a well organized system. The fix moves all properties (present and, hopefully, future) to a new file.
2. Code reusability: The multitude of different kind of information in IRremoteInt.h makes it very hard to reuse the board data base in another project. (Actually, the very name _Int_ suggest that this is not intended.) With the proposed change, including `boarddefs.h` into another project is a sensible thing to do.

... and this exactly what I like to do in [Infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino)


z3t0:
Brilliant idea. But seems to be out of date with the base branch, feel free to rebase it or I will definitely do so when I get a chance

z3t0:
Thanks for your hard work again!

PR #363, opened 2016-09-18
added suport for Motorola IR codes

rabarar:
Added support for the (once included) Motorola IR Codes.


ArminJo:
Missing comments and outdated structure

PR #370, opened 2016-10-14
Added sendSanyo and Samsung36.

CodeMinion:
Samsung36 - IRP notation: {38k,500}<1,-1|1,-3>(9,-9,D:8,S:8,1,-9,E:4,F:8,-68u,~F:8,1,-118)+

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -8,6 +8,7 @@
 // Modified  by Mitra Ardron <mitra@mitra.biz>
 // Added Sanyo and Mitsubishi controllers
 // Modified Sony to spot the repeat codes that some Sony's send
+// Modified by Frank Hernandez <hernandez.frank@gmail.com> to add sendSanyo
 //
 // Interrupt code based on NECIRrcv by Joe Knapp
 // http://www.arduino.cc/cgi-bin/yabb2/YaBB.pl?num=1210243556
--- a/IRremote.h
+++ b/IRremote.h
@@ -59,7 +59,7 @@
 #define SEND_LG              1
 
 #define DECODE_SANYO         1
-#define SEND_SANYO           0 // NOT WRITTEN
+#define SEND_SANYO           1
 
 #define DECODE_MITSUBISHI    1
 #define SEND_MITSUBISHI      0 // NOT WRITTEN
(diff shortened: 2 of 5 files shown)


srblank03:
Hi, I tried this code whit a SANYO CRT tv, and works great!
Thanks!

z3t0:
@srblank03 thanks I have noted that this has been tested and is scheduled to be added to the next release.

bengtmartensson:
The "-68u" in the IRP is definitely moot, see [this](http://www.hifi-remote.com/forums/viewtopic.php?t=100280) (I am "Barf" there.)

Likely some manufacturer wrote an implementation that due to some computational delay (or just sloppy programming) had an extra 68 micro delay, and someone reverse engineered it just too well... It makes absolutely no sense, in particular since 68micros is well within the error tolerances.

jill880829:
Hi. I am trying to decode the signal of Sanyo air conditioner controller.
I would like to know how to use the function sendSanyo( address, data ) in the your library?
What are the variable address and data if I only get the decoded results which is showed as the type NEC?
For example, I get the result "0x702C0200" shown as type NEC, what should I send in sendSanyo() ?

CodeMinion:
Hello @jill880829 ,

If the Sanyo AC unit you have in fact uses the NEC protocol in principle you should be able to use the sendSanyo method though I don't believe this implementation has been tested outside of controlling a TV. 

The **address** in the parameters is the address of the receiving device (in your case, the AC unit). This address is expressed in 16 bits where the first 8 bits is the address of the device and the next 8 bits is the logical inverse of the address. 
In the case of a Sanyo TV the address you might pass is 0x1CE3. Where 0x1C  (0001 1011) is the address and 0xE3 (1110 0100) is its inverse.   

The **data** in the parameters is the command to send to the device. It follows the same format as the address in the sense that this is a 16 bits value where the first 8 bits is the command and the next 8 bits is the logical inverse of the command. 
In the same case of a Sanyo TV above you might pass a value of 0x48B7 for the power off command where 0x48 (0100 1000) is the command and 0xB7 (1011 0111) is its inverse. 

**Note:** You might be able to find your device here: http://lirc-remotes.sourceforge.net/remotes-table.html which might save you the trouble of decoding the values yourself. 

For example, the Sanyo TV in this comment can be found here: https://sourceforge.net/p/lirc-remotes/code/ci/master/tree/remotes/sanyo/TV2.lircd.conf
Where address is the **pre_data** field and the commands are the **codes**. 

The result of "0x702C0200" you are getting from the decode does not seem right this could mean that either: 

1. The decode method might not have been able to decode it properly.
2. The Sanyo AC unit might not be using NEC.

Have you tried using the decodeSanyo method in the library?

ArminJo:
Too much conflicts and the NEC standard encoding was recently added.

Sachintermux:
> 
> 
> Hi, I tried this code whit a SANYO CRT tv, and works great!
> Thanks!

how many bits long your ir signal  my 42 bits long signal and i tried to send but its not work can you help me??

Issue #318, opened 2016-06-16
Arduino Zero/MKR1000 (ARM Cortex-M0+ based) support


q2dg:
I suppose the answer will be the same than here (https://github.com/z3t0/Arduino-IRremote/issues/313) but maybe I'm lucky...
Thanks!!

z3t0:
Hah you're in luck. I am currently actually working on this in cooperation with another community member. 

Thanks, 
Rafi Khan

q2dg:
Yeah!! :-D
Thanks!!

digitalhen:
Hi guys,

Just wondering if there's any progress on this

z3t0:
Sorry, I have been quite busy for a while. But I will be getting to this at some point soon.

trodderus:
I am very keen for MKR1000 support also!

Cheers

z3t0:
Hi just wanted to let you know I am currently stuck with life and so wont be contributing anything to this project for a while, but if anyone does get around to this I am monitoring PRs all the time and will merge them

codehArt:
Hello @z3t0  ,
I want to contribute, if you can guide me, I might help myself.

bengtmartensson:
[This article](https://forum.arduino.cc/index.php?topic=344552.0) can probably serve as a starting point.

"It's left as an exercise for the reader to weave the ZERO functions in to the generic IRremote library"

codehArt:
Hello,
 @bengtmartensson  thanks for your guidance,
However I found all configurations of particular board in IRRemoteInt.h file ,
but unfortunately I don't know which kind of data should be added in  IRremoteInt.h file for arduino mkr 1000 or zero, so I can use this library .

 even I already found where to add mkr1000 board in boards configuration but I am not succeeded to found timer config for arduino mkr1000 32 bit timer
and there is another problem of mkr that it is 32 bit controller so as per my assumption we have to create another 32 bit timer.

agn0stico:
Hi guys.
I am also very interested in this platform support. Has there been any progress by any chance?
I tried to play around a bit with the code in the article that **@bengtmartensson** mentioned. I moved some code of the prototype sketch into _IRemoteInt.h_ to configure timer and PWM, but rather than getting it to compile for the MKR1000 board I didn't have luck getting it to work.

agn0stico:
[...]

PR #379, opened 2016-11-13
[add] preliminary support for Arduino MKR1000

agn0stico:
@z3t0, I was almost forgetting about this pull request. There you go.

### DIFF
--- a/IRremoteInt.h
+++ b/IRremoteInt.h
@@ -87,6 +87,11 @@ EXTERN  volatile irparams_t  irparams;
 #	define BLINKLED_ON()   (PORTD |= B00000001)
 #	define BLINKLED_OFF()  (PORTD &= B11111110)
 
+#elif defined(__SAMD21G18A__) || defined(__SAMD21E17A__)
+#	define BLINKLED        6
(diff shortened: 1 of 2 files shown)


IRremoteInt.h
```diff
@@ -87,6 +87,11 @@ EXTERN  volatile irparams_t  irparams;
 #	define BLINKLED_ON()   (PORTD |= B00000001)
 #	define BLINKLED_OFF()  (PORTD &= B11111110)
```
z3t0: This is an example of flipping bits to achieve the same result as digitalWrite but without being as costly

AnalysIR: However, digitalWrite will be more portable between physical Arduino-ish platforms....and will not be as big an issue on faster MCUs.

I suspect it isn't a big issue for 99% of existing users & advanced users can always fine tune for performance where needed.

Adminius: https://github.com/z3t0/Arduino-IRremote/issues/318#issuecomment-285298447

bengtmartensson:
Please close. Not based on current code.

Issue #380, opened 2016-11-14
Sharp message is split into 3 messages


buestad:
I am working on a project that uses the Sharp protocol. The decodeSharp function is missing in the ir_sharp.cpp as of today, but I have made it in my fork. 
https://github.com/buestad/Arduino-IRremote

My problem is that the message from one command from the remote is split into three messages before I read it in the decodeSharp function. One Sharp message contains the data three times. First all bits are normal, then all are inverted and lastly normal again. In between these is a pause of roughly 40ms. 

Is there a way to merge theses three messages together before the decode functions are called?

AnalysIR:
Probably a good idea to post the model of device & remote (also as a comment in the code), as there will be more than one Sharp IR protocol.

It would be useful to post a sample set of RAW signals here for reference.

buestad:
I use a Denon RC-940 remote control. This has a bunch of different protocols I believe but according to the IR code document from Denon (assets.denon.com/documentmaster/us/denon%20master%20ir%20hex.xls) it uses the Sharp protocol on the stuff I am interested in. I have tested using a Denon AVR 1804 reciever. I have some raw data but it is captured using a different library because I think it is a bit more readable. Here is an example from pressing the power on button:

```
Decoded Sharp: adr:02 data:E1

Received: 

OFF 	ON
53472 usec, 200 usec		
820 usec, 200 usec	0	
1820 usec, 220 usec	1
780 usec, 220 usec	0
800 usec, 180 usec	0
840 usec, 160 usec	0

1860 usec, 220 usec	1	
800 usec, 280 usec	0
740 usec, 200 usec	0
780 usec, 240 usec	0
780 usec, 180 usec	0
1840 usec, 220 usec	1	
1800 usec, 280 usec	1		
1760 usec, 160 usec	1

840 usec, 180 usec	0
840 usec, 280 usec	0

42620 usec, 260 usec
760 usec, 240 usec	0
1780 usec, 200 usec	1
820 usec, 240 usec	0
780 usec, 200 usec	0
800 usec, 220 usec	0

800 usec, 180 usec	0
1840 usec, 200 usec	1
1820 usec, 260 usec	1	
1760 usec, 200 usec	1
1800 usec, 220 usec	1
820 usec, 180 usec	0
820 usec, 200 usec	0
800 usec, 260 usec	0

1780 usec, 180 usec	1
1820 usec, 240 usec	1

42680 usec, 220 usec
780 usec, 260 usec	0
[...]

PR #381, opened 2016-11-14
Added missing Sharp decode function 

buestad:
In the ir_sharp.cpp I have added the missing decode function. I have also modified the encode function to match. 
In IRremote.h and irRecv.cpp I have added define and call to the above function.

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -67,7 +67,7 @@
 #define DECODE_DISH          0 // NOT WRITTEN
 #define SEND_DISH            1
 
-#define DECODE_SHARP         0 // NOT WRITTEN
+#define DECODE_SHARP         1 
 #define SEND_SHARP           1
 
 #define DECODE_DENON         1
--- a/irRecv.cpp
+++ b/irRecv.cpp
@@ -85,6 +85,11 @@ int  IRrecv::decode (decode_results *results)
 	if (decodeLegoPowerFunctions(results))  return true ;
 #endif
 
+#if DECODE_SHARP
+  DBG_PRINTLN("Attempting Sharp decode");
+  if (decodeSharp(results))  return true;
+#endif
+
 	// decodeHash returns a hash on any input.
 	// Thus, it needs to be last in the list.
 	// If you add any decodes, add them before this.
--- a/ir_Sharp.cpp
+++ b/ir_Sharp.cpp
@@ -93,10 +93,17 @@ bool  IRrecv::decodeSharp(decode_results *results)
 	unsigned long  data = 0;  // Somewhere to build our data
 	unsigned long  lastData = 0;  // Somewhere to store last data
 	int            offset = 1;  //skip long space
-
-
-	// Check we have the right amount of data  
-	if (irparams.rawlen != (SHARP_BITS + 1) * 2)  return false;
+	int			   loops = 1; //number of bursts
+	
+	// Check we have the right amount of data
+	// Either one burst or three where second is inverted 
+	// The setting #define _GAP 5000 in IRremoteInt.h will give one burst and possibly three calls to this function
+	if (irparams.rawlen == (SHARP_BITS + 1) * 2)
+		loops = 1;
+	else if(irparams.rawlen == (SHARP_BITS + 1) * 2 * 3)
+		loops = 3;
+	else
+		return false;
 	
 	// Check the first mark to see if it fits the SHARP_BIT_MARK_RECV length
 	if (!MATCH_MARK(results->rawbuf[offset], SHARP_BIT_MARK_RECV))  return false;
@@ -104,14 +111,16 @@ bool  IRrecv::decodeSharp(decode_results *results)
 	if (!(MATCH_SPACE(results->rawbuf[offset+1], SHARP_ONE_SPACE) || MATCH_SPACE(results->rawbuf[offset + 1], SHARP_ZERO_SPACE)))  return false;
 
 	// Read the bits in
-//	for (int j = 0; j < 3; j++) {
+	for (int j = 0; j < loops; j++) {
+		data = 0;
+		addr = 0;
 		for (int i = 0; i < SHARP_ADDR_BITS; i++) {
 			// Each bit looks like: SHARP_BIT_MARK_RECV + SHARP_ONE_SPACE -> 1
 			//                 or : SHARP_BIT_MARK_RECV + SHARP_ZERO_SPACE -> 0
 			if (!MATCH_MARK(results->rawbuf[offset++], SHARP_BIT_MARK_RECV))  return false;
 			// IR data is big-endian, so we shuffle it in from the right:
-			if (MATCH_SPACE(results->rawbuf[offset], SHARP_ONE_SPACE)) addr += 1<<i ;//   addr = (addr << 1) | 1;
-			else if (MATCH_SPACE(results->rawbuf[offset], SHARP_ZERO_SPACE)) addr = addr;  //addr = (addr << 1) | 0;
+			if (MATCH_SPACE(results->rawbuf[offset], SHARP_ONE_SPACE)) addr += 1<<i ;
+			else if (MATCH_SPACE(results->rawbuf[offset], SHARP_ZERO_SPACE)) addr = addr;
 			else                                                        return false;
 			offset++;
 		}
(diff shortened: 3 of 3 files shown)


SergiyKolesnikov:
@buestad Thanks for the patch! I've tested it with a GA825WJSA remote and it works for a single button press. But if I hold the button it decodes some random codes instead of FFFFFFFF.

buestad:
Hi Sergiy.
I was not aware of the repeat code. Can you explalin a bit more how this should work, or guide me to some documentation which explains?

There Is also an issue with my remote (Denon RC-940) for a single press which returns three codes because it sends first normal, then inverted and lastly normal again. These messages are devided by a gap of 42ms. I have learned that there is a setting in IRremoteInt.h where I can specify the gap between messages. I will try to see what I can do with these issues.

SergiyKolesnikov:
I ran the IRrecvDump example with two other remotes. One remote uses the NEC protocol and if I press and hold a button it sends one code for the button and then keeps sending a special code (FF...) to indicate the repetition. The second remote uses the RC5 protocol and if I press the same button multiple times it alternates between two codes (e.g., 7A0, FA0 for channel up), but if press and hold the button it keeps sending the same code. Each protocol has its own way to encode repetition.

Sharp and Denon seem to use the same protocol (the only difference, which can be ignored, is the Expansion bit: 1 for Sharp and 0 for Denon). [Here is a description of the Denon protocol.](https://www.mikrocontroller.net/articles/IRMP_-_english#DENON) It differs from [the description referenced in ir_Sharp.cpp](http://www.sbprojects.com/knowledge/ir/sharp.php) that you probably used for your implementation. The differences:

* Frame length is 16 bit and not 15. The last bit is a stop bit.
* The gap between the initial and the inverted frame is 65ms and not 40ms.
* The pulse and pause timings are slightly different.
* The expansion bit (the left of the two data construction bits) is 0 and not 1 (this can be ignored, because it is only used to differentiate between Sharp and Denon devices).

And the most important part: "Keyboard Repetition: both frames after 65ms." My interpretation is that a gap of 65ms and than both frames in a sequence (the initial and the inverted frame without a gap in between) indicate a repetition.

As I mentioned already, my Sharp remote works fine with your patch and the IRrecvDump example, but it behaves weird with IRrecvDemo example. This example code prints 2 or 3 identical codes for one key press (similar to your Denon RC-940 problem) and some times it prints sporadic codes. This may be an indication for the incorrect protocol description on sbprojects.com and an explanation for the differences with the Denon protocol from mikrocontroller.net.

I started playing with the IRremote library just yesterday, so that is all I could found out about Sharp protocol so far.

buestad:
I filed [an issue](https://github.com/z3t0/Arduino-IRremote/issues/380) some days ago about the core of the API splitting the three frames (normal-inverted-normal) into three messages. This results in three calls of the decodeSharp function. The suggested solution was to increase the setting #define _GAP 5000 in IRremoteInt.h to 50000. This solves the problem for a single button press, but does not work for press and hold, since the gap between the three frames and between each message is the same. Anyhow, I have updated my fork...

In the same bug I have included a raw dump of one message from my remote. As you can see the gaps I have is 42ms, not 65 like the Denon protocol suggests. The exp. bit is 0 and then inverted in the middle frame which should sugest a Denon protocol? 

It would be interesting to see a raw dump from @SergiyKolesnikov 's Sharp remote using IRrecvDump.ino
My Denon RC-940 dump from IRrecvDump.ino looks like this:
```
Decoded Sharp: Address: 2 Data: E1 (15 bits)
Raw (96): 250 -800 250 -1800 200 -850 250 -800 250 -750 250 -1850 250 -800 150 -850 300 -750 250 -800 150 -1900 250 -1800 200 -1850 250 -800 250 -800 150 -43550 200 -800 300 -1750 300 -750 200 -850 300 -750 250 -750 250 -1850 250 -1800 200 -1850 250 -1800 300 -750 200 -850 250 -800 200 -1850 250 -1800 250 -43450 250 -800 200 -1850 250 -750 300 -750 200 -850 250 -1800 300 -750 250 -800 250 -750 200 -850 300 -1750 300 -1800 200 -1850 250 -800 200 -800 300 
```

Press and hold:
```
Unknown encoding: 652C70AB (32 bits)
Raw (101): 250 -800 150 -1900 250 -750 300 -750 200 -850 250 -800 250 -1800 250 -1800 250 -1850 150 -1900 250 -750 250 -800 250 -800 250 -1800 200 -1850 300 -43400 250 -800 250 -1800 200 -850 250 -750 300 -750 200 -1850 300 -750 250 -800 250 -800 250 -750 200 -1900 250 -1800 250 -1800 250 -800 250 -800 150 -43550 250 -750 250 -1800 300 -750 200 -850 250 -750 300 -750 200 -1850 300 -1800 250 -1800 250 -1800 250 -800 200 -850 250 -800 200 -1850 250 -1800 250 -43450 250 -800 250 -1800 
```

SergiyKolesnikov:
Hi @buestad, I took another approach and just ignored the inverted and the first repeated frame. The ignored signals are decoded as unknown by IRrecvDump.ino, but it is not a problem, because one have to check results.decode_type anyway.

The push-and-hold is recognized by the pauses between frames. If it is < 50ms then it is a repeated frame and not a subsequent button press. The repeated frames are decoded as 0xFFFFFFFF.

I adjusted the bit mark timings, such that it should work with your Denon remote too.

My Sharp GA825WJSA dump from IRrecvDump.ino for pressing key 1:  
**Single press**
```
1
Decoded Sharp - Address: 1 Command: 1 (13 bits)
Raw (32): 300 -1750 400 -600 300 -750 250 -750 300 -700 250 -1750 300 -750 250 -800 250 -750 250 -750 250 -750 250 -750 300 -700 250 -1800 300 -700 250 
```
**Push and hold**
```
1
Decoded Sharp - Address: 1 Command: 1 (13 bits)
Raw (32): 300 -1750 300 -750 250 -750 250 -750 300 -700 300 -1750 300 -750 250 -750 250 -750 250 -750 250 -750 300 -700 300 -700 300 -1750 250 -800 250 
A1F51F21 // An ignored inverted frame
Unknown encoding: A1F51F21 (32 bits)
Raw (32): 250 -1750 300 -750 250 -750 250 -750 250 -800 250 -750 250 -1750 300 -1700 300 -1800 250 -1750 300 -1750 300 -1750 300 -1750 300 -700 250 -1750 350 
885B30D5 // The first ignored repeated frame
Unknown encoding: 885B30D5 (32 bits)
Raw (32): 300 -1750 300 -700 300 -750 250 -750 250 -700 300 -1750 300 -750 250 -750 250 -750 250 -750 300 -700 300 -750 250 -700 300 -1750 300 -750 250 
A1F51F21 // The second repeated frame was missed while processing the inverted and repeated frames, that is why here comes another inverted frame.
Unknown encoding: A1F51F21 (32 bits)
Raw (32): 250 -1750 300 -750 250 -750 250 -750 250 -800 250 -700 300 -1750 300 -1700 300 -1800 250 -1750 300 -1750 300 -1750 300 -1750 300 -700 250 -1750 350 
A1F51F22 // The third repeated frame was missed while processing the inverted and the first repeated frame, that is why here comes another inverted frame.
Unknown encoding: A1F51F22 (32 bits)
Raw (32): 300 -1750 250 -800 250 -700 300 -750 250 -750 250 -700 300 -1750 300 -1750 300 -1750 300 -1750 250 -1750 300 -1750 300 -1750 300 -750 250 -1750 300 
FFFFFFFF // The fourth repeated frame is recognized correctly
Decoded Sharp - Address: 1 Command: FFFFFFFF (0 bits)
Raw (32): 300 -1750 300 -750 250 -750 250 -800 250 -700 300 -1750 250 -750 250 -800 250 -750 250 -700 300 -750 250 -750 250 -750 300 -1750 250 -750 300 
FFFFFFFF // The fourth repeated frame is recognized correctly
Decoded Sharp - Address: 1 Command: FFFFFFFF (0 bits)
Raw (32): 300 -1700 300 -750 250 -800 250 -750 250 -700 300 -1750 250 -800 250 -750 250 -750 250 -750 250 -800 250 -700 300 -750 250 -1750 300 -750 250 
```

buestad:
HI @SergiyKolesnikov 

Merry Christmas!
I have not had time to do much work on the IRremote code the last weeks, but I tested your code now and it works 50% of the time. Probably just some small adjustment on the timings. I have decided not to do any more work on this Sharp protocol since I am happy with my project for now.

Issue #14, opened 2012-08-24
about RC5 second bit and RC6 mode6


lxbzmy:
Hi author,
First of all, this a good script and I like it.
 After  researched  many specification about RC5 and RC6 (gogole pronto.pdf),
After researching many specifications about RC5 and RC6 I found [this](http://www.majority.nl/files/pronto.pdf ?)

I found that your RC5 RC6  function need patch to support RC5x and RC6A.

This is my patched version:

``` c
/**
* Note: first bit must be a one (start bit)
* first bit always one, second bit may 0 or 1(pronto.pdf) 0 means command bit will add 64 
*/
void IRsend::sendRC5(unsigned long data, char nbits)
{
  enableIROut(36);
  data = data << (32 - nbits);
 /*  space(RC5_T1);
  mark(RC5_T1); // First start bit
  space(RC5_T1); // Second start bit
  mark(RC5_T1); // Second start bit */
  for (int i = 0; i < nbits; i++) {
    if (data & TOPBIT) {
      space(RC5_T1); // 1 is space, then mark
      mark(RC5_T1);
    } 
    else {
      mark(RC5_T1);
      space(RC5_T1);
    }
    data <<= 1;
  }
  space(0); // Turn off at end
}



/** according rc6 mode 6 data bits may be 32bit ,total bits will be 36bit(3 mode bit 1 toggle),this is over flow long data type
 * Caller needs to take care of flipping the toggle bit
 * @param mode toggle bit(0 or 1) mode bit is locked to 110
 * @param data unsigned 32 bit int
 * @param nbits 24 or 32. see pronto.pdf RC6A.
**/
void IRsend::sendRC6(char head,unsigned long data, char nbits)
{
  enableIROut(36);
  data = data << (32 - nbits);
  mark(RC6_HDR_MARK);
  space(RC6_HDR_SPACE);
  mark(RC6_T1); // start bit
  space(RC6_T1);
  //begin mode bit

  //begin data bit
  int t;
  head <<=4;//only need lower 4bit
  for (int i = 0; i < 3; i++) {
    if (head & 0x80) {
      mark(RC6_T1);
      space(RC6_T1);
    } 
    else {
      space(RC6_T1);
      mark(RC6_T1);
    }
    head <<= 1;
  }
  //tg bit
  if (head & 0x80) {
      mark(RC6_T2);
      space(RC6_T2);
  }else {
      space(RC6_T2);
      mark(RC6_T2);
  }
  //loop 
[...]

PR #382, opened 2016-11-24
Added support for sending RC5x commands

mariusoei:
RC5x commands have a 2-cycle pause after the device ID, so they can't be
sent with the regular RC5 protocol.

This commit adds an sendRC5x function that accepts device ID, command and extension and sends the appropriate IR signal.

I have tested this with the modified IRRemote library for the ESP8266 and a Marantz amplifier.

This commit doesn't support decoding for RC5x because I don't have the means to test this.

See discussion at https://github.com/z3t0/Arduino-IRremote/issues/14

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -31,6 +31,8 @@
 #define DECODE_RC5           1
 #define SEND_RC5             1
 
+#define SEND_RC5X            1
+
 #define DECODE_RC6           1
 #define SEND_RC6             1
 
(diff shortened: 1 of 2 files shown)


nelu22:
Hello, can you help me with an sketch for send RC5X command. I want to learn an Marantz remote some commands.
Thankyou!

mariusoei:
Yeah, what specifically do you need help with?

nelu22:
When i try to use RC5x command in arduino i ghet error from compiler. I am not expert in programming .
Please help me with a little routine how to send one RC5x command on arduino.

mariusoei:
Did you pull this branch? It's not pulled into the master branch yet.

nelu22:
Thank you!

mariusoei:
I believe I have been mistaken here. The RC5x-Protocol is different from what Marantz uses/used(?) in their devices. They seem to call it the "Marantz Extended Data Word", as described [here](http://lirc.10951.n7.nabble.com/Marantz-RC5-22-bits-Extend-Data-Word-possible-with-lircd-conf-semantic-td9784.html) and [here](https://sourceforge.net/p/lirc/mailman/lirc-list/thread/1361951809168-9801.post%40n7.nabble.com/#msg30537120).

I'm closing this pull request because what I implemented wasn't actually RC5x (which is now part of the library anyway).
I might open another one with the correct name at some point.

PR #398, opened 2016-12-12
Update IRrecord.ino

galgolan:
fixed IRRecord example for JVC


z3t0:
@galgolan thank you for your contribution. The PR was out of sync so I just created #406 which is in sync with master. Can you please take a look there to see if it's alright?

PR #403, opened 2017-01-17
Added Supported Sensors

isantolin:


z3t0:
@isantolin I do not have those sensors to test with so if you could provide some sources to confirm with that would be great. For example, did you test those yourself? Or maybe other users did.

z3t0:
@bengtmartensson I agree that seems like a better idea.

Any thoughts @AnalysIR @isantolin

isantolin:
@z3t0  
> I do not have those sensors to test with so if you could provide some sources to confirm with that would be great. For example, did you test those yourself? Or maybe other users did.

Yes, i did the test and has the same usage than TSOP

@bengtmartensson @AnalysIR 
> Better would be to state something like "(all) demodulating sensors with inverted output, supply voltage compatible with the board, and suitable modulation frequency are supported". Creating and maintaining an explicit list creates more problems and misunderstandings than it solves.

In my personal opinion, when i tried to use the library, i cant found any concrete information about compatibility of IRM8601, issues or usage examples and i did some "trial-and-error" tests to do work with that library. This short list can save a lot of time for the users

z3t0:
> This short list can save a lot of time for the users

That's a good point

bengtmartensson:
I can possibly add that IRM8601 IMHO is not a good selection, because
* it is end-of-life. Neither mouser not degikey carries delivers it.
* It is not 3.3V compatible.

AnalysIR:
> Better would be to state something like "(all) demodulating sensors with inverted output, supply voltage compatible with the board, and suitable modulation frequency are supported". Creating and maintaining an explicit list creates more problems and misunderstandings than it solves.

I think Bengt's suggestion is better than listing off individual receivers. At least with Vishay receivers they tend to reference the IR protocols targeted in the datasheet. So maybe adding a note to the above to refer to the datasheet for compatability with individual protocols and to avoid any receiver where the supplier does not provide a datasheet **and** protocol list.

Personally, I would also highlight that the extremely cheap VS1838 receivers & ones that are sold with similar names [should be avoided where possible](https://www.analysir.com/blog/2014/12/08/infrared-receiver-showdown-tsop34438-vs-vs1838b-winner-revealed/).

To cover situations where users harvest receivers from old TVs etc, then it is appropriate to try it & if it works, great!. If it doesn't work, then buy one based on its datasheet. In an emergency, you can also make [your own IR receiver from an IR LED](https://www.analysir.com/blog/2014/05/28/poor-makers-infrared-receiver-2/).

When it comes to longer AC signals, it is a different matter or special case - as they aren't typically mentioned in the data sheet. However, the section titled
**SUITABLE DATA FORMAT**
usually gives a good indication if a receiver can handle longer AC signals. My current favourite for AC signals is [TSOP34438](http://www.vishay.com/doc?82489). Again, if there is no such information available in the datasheet, then users should avoid that receiver.

In conclusion, like for any electronic component...check the datasheet **first**!.

isantolin:
@AnalysIR I added the Datasheets URL for components

isantolin:
@z3t0 you can Accept or Reject this PR?

z3t0:
@analysir is this good to merge or do you think changes need to be made? You're more of an expert on the electronics involved than myself :)

AnalysIR:
@z3t0 Although I appreciate the efforts by @isantolin , I suggest not adding in a list of receivers as most IR receivers are 'supported' by the library.

**Reason**: based on the comments provided by @bengtmartensson and myself above. For example, there are only a few IR receivers not supported by the library. It doesn't seem to make sense to  just list all products manufactured by Vishay and any other manufacturer etc.  For example Vishay don't make bad receivers, they make **many** different receivers for specific uses, footprints, pin-outs, protocols, frequencies etc (all explained in the individual data-sheets).

z3t0:
On the advice of @AnalysIR I am deciding it would be advisable to not add a list of supported receivers.

Issue #375, opened 2016-11-10
Recommend more specific header file name


PaulStoffregen:
Please consider renaming "boarddefs.h" to something more specific, like "IRremoteBoarddefs.h".

Arduino indexes all .h files.  If another library also uses this generic name, a library matching conflict can occur, where Arduino may infer the use of the wrong library.  These issues are not common, but when they do strike the result is terribly confusing.

Changing the name to be IRremote specific is a simple and easy thing.

bengtmartensson:
If this is an issue (I am not saying or implying it isn't), this should go into some general library guide lines. (Possibly there is already one?)  Where is the indexing described?

But them, are the existing file names specific enough?

z3t0:
Are there any other files that should be renamed?

>  Where is the indexing documented or described?

Perhaps as @bengtmartensson has suggested there is somewhere we can check?

z3t0:
One way to move forward with this?

> Finally, I noticed the recent update with a boarddefs.h file. You might consider putting this into a sub directory with one file per 'board/MCU' supported, along with conditional compiles to include the correct board file. I believe this will make it easier for newcomers to grasp the details vs one increasingly large file.

Quoted from: @AnalysIR

bengtmartensson:
I opened [this thread](http://forum.arduino.cc/index.php?topic=439926.0) in the Arduino forum.

    One way to move forward with this?...

It would rather amplify the problem,  with files like, say, boarddefs_uno.h

bengtmartensson:
I tried to understand what really happens.

When calling the compiler, it is given a list of include directories to search for include files (-I...). This can potentially cause problem if a file is found in several of these. However, boarddefs.h is included from IRremoteInt.h by
     
    #include "boarddefs.h"
not by

     #include <boarddefs.h>

This means that it is searched in the same directory as IRremoteInt.h, so that if the correct version of IRremoteInt.h is found, the correct version of boarddefs.h will be used, also if there are other versions in other include directories. I have even tried it. boarddefs.h is a private, internal file that should not be directly included by any file not belonging to the library. So, IRremote is immune to other libraries using boarddefs.h.

The only thing that can bite is if another library decides to introduce a file called boarddefs.h, and to make it public in the sense of recommending users to `#include <...>` it (as opposed to `#include "..."`). Under some circumstances, it may find the  IRremote file. 

It is my opinion that, since "our file" does not break any published guidelines, it is not a "bug", and no change is necessary. 

[...]

PR #405, opened 2017-01-21
renamed "boarddefs.h" to "IRremoteBoardDefs.h"

z3t0:
#375 

Waiting for confirmation by at least another contributor before merge into master

### DIFF
--- a/IRremoteBoardDefs.h
+++ b/IRremoteBoardDefs.h
@@ -17,8 +17,8 @@
 // Whynter A/C ARC-110WD added by Francesco Meschia
 //******************************************************************************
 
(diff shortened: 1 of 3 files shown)


z3t0:
need to update the branch

bengtmartensson:
Since the said file has been relocated to a subdirectory (`private`) it is not exported, the reason for the renaming is no more existing.  I recommend that this PR is not merged.

ArminJo:
Since the boarddefs.h is only referenced once in IRremoteInt.h, I would propose to do the renaming,
It is faster done, than writing this comment.

PR #413, opened 2017-02-02
Added support for the "Commodore Amiga CDTV" IR protocol

salocinx:
The CDTV uses a 40kHz tranceiver/receiver. My protocol implementation works very robust including checksum validation (cp. lower 12-bit with inverted higher 12-bit). The CDTV sends/receives 24-bits in total and uses a repeat-signal (#FFFFFF) when buttons are hold (like the NEC protocol).

PR #415, opened 2017-02-07
Salocinx master

z3t0:
#413 

@salocinx

I needed to recreate the PR to add the changelog commit. I will give the code another look in the next few days and then merge within the week.

Thank you very much! Your contribution is greatly appreciated. If possible, could you summarize in a sentence or two any difficulties you may have had in creating this protocol spec? Just something to add to the wiki for any future coders that want to add their own protocols.


z3t0:
The branch has become out of date again, should be a trivial fix. Also the new protocol needs to be added to the relevant example sketches. And we need documentation to explain the protocol specs

ArminJo:
Please rebase the code if the protocol is still required.

PR #433, opened 2017-04-05
Revert "Added ESP32 IR receive support (IRsend not implemented yet)."

z3t0:
Reverts z3t0/Arduino-IRremote#425


ArminJo:
ESP32 support is now implemented

Issue #386, opened 2016-12-04
PSA : Accepting Feature Requests


z3t0:
Hi All,

### Announcement

First of all I apologize for not being active in the community due to my other commitments.

~~However! I will be spending a whole week during the Christmas break to completely revamp this library.~~

EDIT: This has been pushed back BUT is till on my todo list so keep adding suggestions and I will work on them as time permits, I apologise for the inconvenience

I will be accepting any and all requests, regardless of their nature. There is however one caveat. If you want me to add support for a board not currently supported then you may have to spend some money shipping one to me. But ask me first as I do have quite a collection and so might already have access to the board.

### Steps
1. Create an Issue **detailing** your request. Please provide as much detail as possible.
2. Tag me by typing "@z3t0" somewhere in the issue
3. Tag this issue by typing "#386"
4. Await patiently and check below to see once your issue has been added to the list to be worked upon.

### Issues to be worked on:
*if your request is in this list then it has been brought to my attention and I will work on it, if not please ensure you follow the steps above!*

- [ ] [Write Tutorials and Make Videos for the End User](https://github.com/z3t0/Arduino-IRremote/issues/387)
- [ ] [Fix calculation of pause length in LEGO PF protocol](https://github.com/z3t0/Arduino-IRremote/pulls/385)
- [ ] [Write Tutorials on Development](https://github.com/z3t0/Arduino-IRremote/issues/388)
- [ ] [Library Goals Statement](https://github.com/z3t0/Arduino-IRremote/issues/389)
- [ ] [Recommend more specific header file name](https://github.com/z3t0/Arduino-IRremote/issues/375)
- [ ] [Add support for ESP2866](https://github.com/z3t0/Arduino-IRremote/issues/400)

z3t0:
@AnalysIR @bengtmartensson @shirriff @joshuajnoble @pcoughlin @csBlueChip 

Just thought I'd tag some of the main contributors! Sorry for the spam 😄

AnalysIR:
A Tutorial on how to add a new Board, might be very useful, with one section devoted to Board suppliers/makers.

Similarly, a Tutorial on adding a new protocol....

It seems the above 2 points would lead to an ever increasing number of supported platforms/protocols.

Of course, one the the biggest barriers to getting 'code' contributions is the learning curve for making contributions to GitHub (PRs etc). I have made a few, but I know I couldn't explain the process properly. Maybe this already exists elsewhere and just needs to be linked?.

[...]

PR #437, opened 2017-04-12
Cleanup of hardware dependencies. Merge in SAM support. 

bengtmartensson:
This PR attempts to clean up the low-level hardware dependencies. This problem was made glaring by  #425 and its implementation by @marcmerlin. Later, the changes for SAM boards (see #390), first by @MrBryonMiller, later fixed by @ladyada, https://github.com/adafruit/Arduino-IRremote. These changes have been "cleaned" in the same sense.

All explicit hardware dependencies are now either in `boarddefs.h`, or in a board-specific file. Conditional compilation in the other files are now exclusively governed by "high-level" properties. Soft carrier (which sometimes is called "bit-banging") have been implemented hardware-independent.


marcmerlin:
First, I'm just going to say is thank you for the architecture clean up and improvements and doing a better job than me abstracting all this.

Second, since you also had a look at the receiving code, I was wondering why it was based on a timer that fires no matter what, vs a pin interrupt that would only fire each time the IR signal toggles from low to high or high to low. Is it because on some AVR chips it's just difficult to do this on any pin while on newer chips per pin interrupts are trivial?
If so, I can see that having 2 versions of the IR receive code, one that is designed to work on pin level change and counting time between each interrupt, vs the one that fires on interval and samples the pin, would probably not be very desirable since it would be 2 different code bases.

marcmerlin:
By the way, does ESP32 in that branch need re-testing? If so let me know and I'll happily do it.

AnalysIR:
@marcmerlin 

> I was wondering why it was based on a timer that fires no matter what

I suspect the original design decision was related to the relatively limited resources available on 8-bit AVRs and that reserving interrupts in addition to 1 timer would reduce/limit the wider application/adaptation of the library (on 8-bit MCUs). This would still be valid today for AVRs and will continue to be a 'conflict' as more and more lower cost 32 bit MCUs are released. You might of course design it differently, if only focused on 32 bit MCUs with lots of flash, RAM & peripherals available or as a stand-alone 8-bit application.

It would be interesting to know the trend/quantum of active users of 8-bit vs 32-bit MCUs over time. I suspect it is still largely 8-bit in Arduino-land(guess?), particularly with beginners & non-experts. @z3t0 ?

marcmerlin:
@AnalysIR I'm confused. Are you saying it's less expensive on a limited resource chip to interrupt based on time no matter what, than to have a pin interrupt that only fires if the pin actually changes?
I thought it was exactly the other way around, which is why I was surprised about the design and I was thinking there must be another limitation that escapes me (like most likely the limitations in getting interrupts on an specific pin on some AVRs)

AnalysIR:
> Are you saying it's less expensive on a limited resource chip to interrupt based on time no matter what, than to have a pin interrupt that only fires if the pin actually changes?

No. Simply when creating a library for general use along with other 3rd party libraries...all on a resource limited platform....it makes less sense to 'reserve/use-up' multiple resources. Essentially for AVRs the library uses one timer (for Rx &Tx/PWM). If it was designed with interrupts, then it would still use up 1 timer for TxPWM plus the interrupt/pin. 

If it was the only functionality running on the MCU, then not a big problem. However, if you want it to work alongside other libraries & user applications, then it is a good idea for the library to use up as few resources as possible - on an 8-bit platform.

z3t0:
I think the point being made is that the library tries to leave as many resources unused as possible so that the rest of the application can use it. 

Though this is something worth investigating. 

#386

marcmerlin:
@AnalysIR ok, maybe I just don't really understand how the code works.
When I was talking about using a pin interrupt for IR reading, I meant only that. I thought you could just track the input pin state change and how much time elapsed between those changes? (and not have a timer at all anymore). In other words, you'd only have a single on demand interrupt and free up a timer.
Isn't that a win on all platforms assuming it can indeed work that way?

AnalysIR:
> In other words, you'd only have a single on demand interrupt and free up a timer.

The additional timer is still required to generate the Carrier PWM for IR Tx.(generating the carrier without a dedicated Timer using bit-banging will vary by MCU/XTAL/Compiler/Optimisation and would be messy to manage/support for a general purpose library). All in all, the original design choice has stood the test of time very well for 8-bit MCUs. 32 bit platforms complicates things because of the relative abundance of resources. It is also important to note that standard IR receivers can distort mark/space timings by up to +/-200uSecs.

For custom & stand-alone applications, interrupts are clearly the better way to go on all platforms....although I remember looking at the new ESP32 datasheet and it seems to have some new dedicated IR peripheral built in, which on first read seemed to be overkill, as typical interrupts/PWM/Timers seem to serve all IR needs. I have ordered some ESP32s to investigate this  new IR peripheral.

Essentially, it all boils down to making design decisions, based on the future 'goals' of the library while also taking the existing user base into consideration. I believe @z3t0 is looking into documenting the Goals and once that is in place getting answers to questions like yours should be more straightforward.

marcmerlin:
@AnalysIR thank you for explaining, I totally missed out on the fact that this library uses a timer for sending too (which I haven't used at all). So if you're only receiving, you're losing out by using a timer over a pin interrupt, but if you're doing both, you've already lost the timer, so you might as well use it for both, makes sense.
For ESP32, you are correct, it comes with an RMT which can be used to basically make precisely timed waves of any kind (8 channels). It was originally meant for IR, but is also being used for other things like neopixels for instance. This is also because of this that I didn't try to port the current sending code to ESP32 (although I also had no use for it), because RMT is the better way to do it and would require new, separate code.

marcmerlin:
@bengtmartensson I just tested this on ESP32, it works fine. You're just missing one line for it to compile ok and after that it's ok:
```
--- a/boarddefs.h
+++ b/boarddefs.h
@@ -607,6 +607,7 @@
 // https://github.com/ExploreEmbedded/ESP32_RMT
 #elif defined(IR_TIMER_USE_ESP32)
 
+#define TIMER_RESET
 #ifdef ISR
 #      undef ISR
 #endif
```

bengtmartensson:
@marcmerlin , @AnalysIR , @z3t0 
Discussing interruptdriven IR reception is, although interesting and important in its own right, here offtopic. I have thumb-down-ed OT responses. Lets keep this discussion on the topic of the PR.

AnalysIR:
> @marcmerlin , @AnalysIR , @z3t0
> Discussing interruptdriven IR reception is, although interesting and important in its own right, here offtopic. I have thumb-down-ed OT responses. Lets keep this discussion on the topic of the PR.

Although in general you are of course correct. In this case, I have to disagree as I believe that the 'discussion' was both warranted and worthwhile, particularly given the recent efforts of @marcmerlin on this and related topics. 

I guess as long as PRs & Issues are the only places for discussion, there will always be a certain amount of spill-over like this.

AnalysIR:
Regarding the bit-banging of the IR signal carrier:
It would be useful to verify correct operation against a scope or LA.
This would also be relevant if more than one clock frquency is supported on the platform, in which case the calculations would need to take that into account or at least specify that only one clock speed is supported.

MrBryonMiller:
I started looking at @bengtmartensson  excellent work pulling all of this together.  Thank you! In reviewing the result I was going to make comments in two areas but first I thought I could help out by testing compilation and operation on an AdaFruit Feather M0 (a SAMD processor).  I realized that the SEND_PIN option I initially allowed for (and which @ladyAda kept in her version) had been removed.  Now the pin is hard-coded as pin 3. Unfortunately I cannot find this pin at the headers of the product so I cannot test nor use this PR unless I modify it.  This brings me to my first comment.

1) I understand that it might be consistent with the original library to not allow user definition of the send pin.  But I would maintain that this would be a good time to incorporate this ability.  Especially as new 32 bit processors are added and soft carrier becomes more prevalent it would seem a shame to use an arbitrary assignment. Furthermore if user definition is allowed even for AVR processors it would be a very useful feature to have a run-time error stating that the chosen pin is not supported.  It would be even more useful if the error could state just what pin is supported. (As a new user I found this information difficult to figure out without digging through the library's code) 

2) I saw that @bengtmartensson generalized the soft carrier output by going away from the hardware dependent port assignment I used (and which @ladyAda kept in her version) to using digitalWrite. I'm guessing because of this generalization a need was felt to compensate for the overhead time by using PULSE_CORRECTION_ON and PULSE_CORRECTION_ON.  These are set to 5 and 6 usec respectively.  For a 38kHz pulse, the half period is 13 usec.  So it seems almost half the time is allocated in an open loop manner.  I set out to measure what was actually needed on the 48MHz Feather M0. My software based measurements suggest that if I used the hardware specific method the loop could be exited in less than 1 usec of the correct time.  This is why I felt no compensation was necessary. But if the general digitalWrite is used the loop exit could be delayed by as much as 2.6 usec.  At his level I would agree that compensation should be applied, but rather than 5 or 6 usec the values should be more like 1 and 2. To me this suggests that using the hardware specific method has value in there is less that could go wrong.  But if the more general digitalWrite is used there should be a hardware specific setting for the two correction values.  And no matter which way prevails, as @AnalysIR  suggests, the output must be verified with a 'scope (something I cannot do).

bengtmartensson:
@MrBryonMiller :
Pin assignment: First of all, you can change TIMER_PWM_PIN (line 72 in boarddefs.h) to whatever sensible, (Yes, the name is now "semantically drifted".)  I wanted to keep the API for _this_ level of PR, therefore I did not make it user selectable. But of course, it would be silly to keep it that way in the long run, just as you say. What do you say about (at least for now) having the constructor for IRsend having a default argument?

    IRsend(pin = SEND_PIN)

@z3t0 : how about  renaming `TIMER_PWM_PIN` to `SEND_PIN`? It _is_ the pin used for sending; putting _how_ it is sent into the variable name seems wrong.

Soft carrier (which I consider a better name than "bit banging"): I used an arduino.org M0+, 48kHz. (Sorry for that; I _ordered_ it as "Arduino Zero"). I tested with IRSenddemo (which is supposed to generate a 40kHz carrier), and analyzed the generated signal using a board [like this](http://www.harctoolbox.org/arduino_nano), deploying a non-demodulation TSMP58000 and the [AGirs](https://github.com/bengtmartensson/AGirs) firmware and IrScrutinizer.  (I have high confidence in this configuration.) I then, with parameters like @ladyada's,  measured around 32kHz. So I changed them to the current ones, which gave pretty exactly 40kHz. So, in this sense, it HAS been tested empirically, although not extensively. (No need to test with a 'scope, though.)

Having said that, I am not all convinced about the sanity of the implementation, neither the old one nor mine. It is completely non-robust against (varying) computational delays. I/We should probably try to cook up something better. But, if possible, a hardware carrier (hardware PWM) implementation should normally be preferred over a soft carrier implementation.

z3t0:
> @z3t0 : how about renaming TIMER_PWM_PIN to SEND_PIN? It is the pin used for sending; putting how it is sent into the variable name seems wrong.

Makes sense to me.

z3t0:
> But, if possible, a hardware carrier (hardware PWM) implementation should normally be preferred over a soft carrier implementation.

That is generally my experience as well. Are there any particular benefits for a soft implementation?

MrBryonMiller:
@bengtmartensson 

> you can change TIMER_PWM_PIN to whatever sensible

Yes, I will edit library file so I can give this a try.

> What do you say about (at least for now) having the constructor for IRsend having a default argument? <br> IRsend(pin = SEND_PIN)

Yes! This is the way I originally had it. @ladyada added in the invert argument but later agreed that it was a little too much.

> how about renaming TIMER_PWM_PIN to SEND_PIN?

I know you weren't asking me about my opinion but I thought I'd weigh in.  Only reason to keep the name TIMER_PWM_PIN is it lets you keep the description "semantically drifted".  I **love** that description.

> @z3t0 : Are there any particular benefits for a soft implementation?

I'm embarrassed to admit but the main reason was to get something working for a project I was building without having to dig through more processor specs about it timers and output multiplexers.  Added benefit was it didn't tie up a second timer and it allowed complete output pin flexibility. I'm now wondering if the same timer used during receive couldn't be switched during send to run a PWM (and switched back when receive is re-enabled).  Also whether with the M0's output multiplexers whether broad output pin flexibility is free even with HW PWM.  I'm thinking about digging into this but results could be months away.

Regarding the use of the PULSE_CORRECTIONs terms :  According to my software base measurement, corrections should not be necessary, but @bengtmartensson actual measurements suggest they are needed.  I have to resolve this in my mind.  So I have managed to round up a fairly modern, fairly accurate oscilloscope (as a supplement for my 1970's era Heathkit) and I'm going to be investigating pulse width and frequency accuracies with and without corrections.  Hopefully results to follow in a day or so.

bengtmartensson:
I have checked in a few commits to the fork, addressing all issues raised. The soft carrier code has been completely re-written. I can't believe that I checked in something like the old version.... Now, on an ideal processor, it should generate exact result (which was not the case with the old implementation).  Most importantly, it does not depend on ad-hoc constants. If one computation lags behind, the lags will not add up. The exact duration of the on-part may vary (PULSE_CORRECTION), and thus the duty cycle, but the generated frequency will be exact, and not dependent on "cosmological constants"! (Unless the CPU is overloaded of course...). Short enough to be included:

     void  IRsend::mark (unsigned int time)
     {
     #ifdef USE_SOFT_CARRIER
        unsigned long start = micros();
        unsigned long stop = start + time;
        if (stop < start)
                // Counter wrap-around, happens very seldomly, but CAN happen.
                // Just give up instead of possibly damaging the hardware.
                return;

        unsigned int count = 0U;
        while (micros() < stop) {
                count++;
                unsigned long now = micros();
                int onTime = min(periodOnTime, (int) (stop - now));
                if (onTime > 0) {
                        digitalWrite(sendPin, HIGH);
                        delayMicroseconds((unsigned) onTime);
                }
                digitalWrite(sendPin, LOW);
                unsigned long targetTime = min(start + count * periodTime, stop);
                int timeOff = (int) (targetTime - micros());
                if (timeOff > 0)
                        delayMicroseconds((unsigned) timeOff);
        }
    #else
        TIMER_ENABLE_PWM; // Enable pin 3 PWM output
        if (time > 0) custom_delay_usec(time);
    #endif

Has been tested (somewhat) with the setup described above.

@z3t0 , @MrBryonMiller 
A port to a board only supporting soft carrier is a "bad" port, and a big TODO. Nevertheless, having a sensible, hardware independent soft carrier implementation cannot be a bad thing.

bengtmartensson:
Pushed some more commits. Comments on the last one:

It is often claimed that `digitalWrite()` is slow, and should be replaced by a more efficient (but non-portable) implementation in time critical application. I am very skeptical to "folklore" understanding of efficiency (replacing constant multiplication by shifts being one example pf pure sillyness), but assuming it is correct, the above implementation is sub-optimal. The last commit replaces the digitalWrite-s by macros, which can be defined to something more efficient.

MrBryonMiller:
@bengtmartensson I apologize for taking so long to comment.  I spent a lot of time trying to figure out what your new "mark" code was doing.  Then I spent time testing at the bit level. Then testing at the overall working level.  My results so far:
1. what your new code does: I finally realized the problem you were trying to solve. In my work I was just worried about getting the bit time right.  I see now that you were just as concerned about getting the correct time for the entire mark.  My code, once it decided mark time was not up, proceeded to send a high then a low even if this pushed the overall mark time past its limit.  Your code cuts mark off at very close to the correct time. Well done!  Just two issues - I believe where you are calculating onTime and you are taking the integer part of (stop - now) there is a wrap-around if the mark time is greater than 32 msec. Don't know how likely this is - doesn't happen in the protocols I use but I have read that some AC protocols use some pretty funky numbers.  Second, where you calculate targetTime I believe "start + count * periodTime" can wraparound. Pretty rare case, but possible at the same rate of the case you protected against, and when it happens hangs the routine up for a long time.
2. I looked on my borrowed scope at the bit times. Looked _very_ close to what was expected. Saw a little jitter at the bit level but I think that might be having to do with bits getting cut off when the mark time was up.  I'll work on my scope triggering and see if I can confirm that is what is going on.  Overall mark timing appeared just right.
3. Finally I tried library in a working example.  I am sending NEC codes from one machine, with new library to another.  Here is where I had biggest disappointment. Seems something was messing with transmission about half the time. Everything would work fine for several sends in a row then recpetion would indicate errors in results.  I'm still digging into whether this is a problem in my example or some difficulty in the library.  My test setup and code seems to be very consistent with my library and @ladyada's library but not this one.  I'll try to dig into this and see if I can figure out what is going on

bengtmartensson:
@MrBryonMiller ,
thank you for you comments. 

1. The main thing is to keep the generated _frequency_ correct. Minor jitter in the individual pulses is not an issue, varying and even incorrect frequency _is_, Recall that almost all receiving equipment uses a demodulating TSOP-ie. I would go so far as to say that sending a half period ON, and a half period of OFF is bad coding, guaranteed to produce wrong result if computational delays are > 0.

> I believe where you are calculating onTime and you are taking the integer part of (stop - now) there is a wrap-around if the mark time is greater than 32 msec.

Agreed. (Formally, it is not "integer part" but casting a long to an int.)  (Also, on platforms where int == long, your argument is not valid.) If we just replace that line by

    long onTime = min(periodOnTime, stop - now);

(_not_ unsigned long), do you see any problems?

> where you calculate targetTime I believe "start + count * periodTime" can wraparound

Hmm. We just checked that stop = start + time did not wrap around. Right now it seems that we could check for that stop + periodTime > start, then we should be clear, since count*periodTime <= time. Comments?

3. Your writing here is very imprecise, and I am not sure what to say. Can there be a resource leak?

MrBryonMiller:
> long onTime = min(periodOnTime, stop - now);
(not unsigned long), do you see any problems?

Since periodOnTime is an unsigned int I'm afraid I don't know enough about min nor do I have a firm grasp on C's casting to know for sure if this will work.  I believe so, just not the right person to ask.  I do agree that this may all be moot since the size of int's and long's are the same (at least on my Feather M0).

> Right now it seems that we could check for that stop + periodTime > start, then we should be clear, since count*periodTime <= time. 

Sorry, I'm a little dull in the morning. I'm not sure if you are agreeing with me that there might be a wrap-around or not.  So at the risk of repeating myself I'll explain my concern in a little more detail. In the iteration through the loop that will turn out to be the very last iteration, count has incremented to the point where start + count * periodTime may very well be greater than stop. If this were not the case you would not need to take the min.  So, while start+time did not wrap around, start + count * periodTime may.
Am I making any sense?

> Your writing here is very imprecise, and I am not sure what to say. Can there be a resource leak?

Sorry, it was even earlier when I wrote that.  There could be any number of problems.  I was just letting you know that I cannot consider myself as having tested this until I get this figured out.  After running an errand this morning I plan on spending the rest of the day on this.

boarddefs.h
```diff
@@ -574,4 +639,13 @@
 #	error "Internal code configuration error, no known IR_USE_TIMER# defined\n"
 #endif
 
+// Provide default definitions, portable but possibly slower than necessary.
```
z3t0: I think this is a great idea! Should we add something like 

```c
#warn Defaulting to software implemented send may be imperfect. See SOME_LINK
```

Or some other disclaimer so that while potentially supporting _unsupported_ platforms we also prevent issues from being created for _bad_ signals being produced?

MrBryonMiller:
@bengtmartensson I am sorry to report I cannot get send to work reliably with this library.  I have a a receiver program that just shows what code is being received.  I have a sender program that 50 times in a row sends a specific code using NEC 32 bit protocol.  What I found was I received the proper code 29 times. Receiver said it got a JVC code 10 times and it got an unknown code 11 times.  I consider this very unreliable. 

To make sure I didn't have anything wrong in my test I changed from your library to the one @ladyada made.  With no other changes to my setup nor to my program I received the correct code 49 times out of 50.

I suspect something is funny in the mark function.  To confirm I used your library but replaced the mark code with the following
```
	unsigned long start = micros();
	unsigned long stop = start + time;
	if (stop < start)
		// Counter wrap-around, happens very seldomly, but CAN happen.
		// Just give up instead of possibly damaging the hardware.
		return;
	while (micros() < stop) {
	  digitalWrite(sendPin, HIGH);
	  delayMicroseconds(periodOnTime);
	  digitalWrite(sendPin, LOW);
	  delayMicroseconds(periodTime-periodOnTime);
	}

```
This is very similar to the original bit banging code and I know you don't like it but this also had a very high reliably (unfortunately I don't remember how high)

I cannot figure out what might not be working in your version but I did consider maybe a simplification that maintained you goals might work.  Hence I replaced the mark code with the following

```
	long m = micros();
	long markStop = m+time;
	while ((markStop-m) > 0) {
		digitalWrite(sendPin, HIGH);
		long highStop = m+periodOnTime;
		while (((highStop-m) > 0) && ((markStop-m) > 0))
			m = micros();
		digitalWrite(sendPin, LOW);
		long lowStop = m+periodTime-periodOnTime;
		while (((lowStop-m) > 0) && ((markStop-m) > 0))
			m = micros();
		}

```
Because this uses longs (rather than unsigned longs) with subtractions/additions I **think** it takes care of all wrap-around problems (without having to throw away a mark if it was close to the wrap-around point). It also eliminates any delayMicroseconds problems.  When I tested this version I received 50 out of 50 correctly.

I am curious to know if you had a chance to test your library and what your results might have been.  I am also very curious to hear what you think about my proposed version of mark especially with respect to any wrap around issue I may have missed.  Finally I wonder if you think with my version whether we'd still need a PULSE_CORRECTION  in the periodOnTime calculation?

Thanks so much for your work on this.

z3t0:
Just a disclaimer: sorry for not actively participating in the process of this PR. I am just busy with other things in life, but am actively monitoring this PR and commenting my thoughts when relevant.

boarddefs.h
bengtmartensson: I have some ideas for a slight reorganization of `boarddefs.h`, which I will elaborate in a future issue. Hopefully today.

Generally speaking, I think `#warn` is a bad idea. They just nag the user (who im most cases do not understand them), and dilutes the build output, which is hard to read anyhow, using the Arduino IDE. Also, `#warn` is, as opposed to `#error`, non-standard, i.e. not contained in the C specification. For example, the Microsoft Visual studion C++ compile has it with a completely different syntax. (OK, we are all using something that is based on gcc, so you may argue that the argument is not valid...)

bengtmartensson:
@MrBryonMiller :
There are two issues here. First I address the wrap-around. I claim that the fix I proposed,

> we could check for that stop + periodTime > start,

works. In detail, consider

       if (stop + periodTime < start )
            // Counter wrap-around, happens very seldomly, but CAN happen.
            // Just give up instead of possibly damaging the hardware.
            return;

    unsigned int count = 0U;
    while (micros() < stop) {
            count++;
            unsigned long now = micros();
            int onTime = min(periodOnTime, (int) (stop - now));
            if (onTime > 0) {
                    digitalWrite(sendPin, HIGH);
                    delayMicroseconds((unsigned) onTime);
            }
            digitalWrite(sendPin, LOW);
            unsigned long targetTime = min(start + count * periodTime, stop);
            int timeOff = (int) (targetTime - micros());
            if (timeOff > 0)
                    delayMicroseconds((unsigned) timeOff);
    }

Proof: If we have passed the first `if`, then `stop + periodTime` <= ULONG_MAX (the largest number that can be represented). Denote the value of count at the last iteration by C (after the increment), and the value of micros() at that time by m. At the beginning of the last iteration of the loop, m = micros() >= start + (C - 1)*periodTime. Since it _is_ the last iteration, m < stop. Question: is start + C * periodTime <= ULONG_MAX, in which case we are OK. Putting this together, we have 

start + C * periodTime = start + (C - 1)*periodTime + periodTime <= m + periodTime < stop + periodTime <= ULONG_MAX.

Q.E.D.

MrBryonMiller:
@bengtmartensson 
 > we could check for that stop + periodTime > start,

But isn't it possible, even if we pass the first if,  that stop is a very large number, nearly ULONG_MAX.  And if so, stop + periodTime may wraparound and become a relatively small number surely much less than start So we can enter the loop in a case where start + count * periodTime can also wrap?

bengtmartensson:
@MrBryonMiller 
Possibly I do not understand your objection/question. Note that the initial "if" is now if (stop + periodTime < start)  (which amounts to stop + periodTime <= UMAX_LONG). My posting proves, with mathematical precision, that in this case, start + count * periodTime <= UMAX_LONG.

MrBryonMiller:
@bengtmartensson 
I apologize. I now see what you are saying. I agree your code with the new if is protected from wrap arounds.

I still await your comments about the difficulty I am having getting reliable results.  And my proposed modifications that, at least for me, were reliable and which don't require throwing away a mark when micros() gets close to wrap-around.

AnalysIR:
Not really sure if this code is intended for only 32-bit or all supported platforms.

If it is also targeted at standard 16Mhz AVR Arduinos, the micros() function only returns time in 4 uSecs increments.
So trying to generate a 56kHz carrier(17.8uSec period) with a resolution of 4 uSecs might be a challenge with this method. _(based on some quick calculations could land anywhere from 45-72kHz, with a slightly smaller range at other frequencies - taking the 4uSecs into account. Code delays might be worse??)._

Maybe averaging would improve things and produce a workable waveform?

Just ignore if not relevant.

bengtmartensson:
@MrBryonMiller ;
> I still await your comments about the difficulty I am having getting reliable results. 

working on it... Can you provide a hardware-pwm implementation in the meantime :-)?

Two important remarks that I forgot to mention: 
1. Note that the if-test condition cannot be improved.
2. The result (overflow cannot occur) holds true also with small computational delays.

@AnalysIR :
It is supposed to be "maximally portable code"; of course it won't work on a "too slow" (whatever that means) processor. I just discovered myself that periodTime = 1000/khz does not distinguish between e.g. 40 and 39 kHz... But most importantly, your argument is a killer for the original code, because there the errors add up -- all in the same direction. The magic word here is using `start + count * periodTime`, which prohibits errors to accumulate.

AnalysIR:
> I just discovered myself that periodTime = 1000/khz does not distinguish between e.g. 40 and 39 kHz

I would usually do something like  `periodTime = (1000+khz/2)/khz` for integer division, if there are enough MCU cycles available.

bengtmartensson:
@AnalysIR 
that is the difference between rounding and truncation, but that is not the issue. (Although you are right in principle; rounding is here more appropriate than truncating) The issue is that
d/d(khz) 1000/khz = -1000/khz^2 , which for khz = 40 turns out to be -0.625 (similar for your formula, it differs only by a constant, and the derivative of a constant...). So a change of khz by dkhz results in a change by -0.625dkhz (approximately). Differently put, that value changes very little  with the parameter.

MrBryonMiller:
@bengtmartensson 
> Can you provide a hardware-pwm implementation in the meantime :-)?

I looked into this a few days ago and immediately ran into the question of which timer is available. We already are using TC3 for receive.  Now I'm thinking, since we never can receive and send at the same time, why don't I try to use TC3 for a PWM during send and switch it back to it's receive function with enableIRIn.  I will pursue that unless anyone weighs in with a show-stopper.

> Note that the if-test condition cannot be improved.

I don't know what you are trying to tell me with this note.  I still believe we don't need any _if_ test at all if we use signed math and let the wrap-around's give us the correct answer (like I did in the code I posted here about one day ago).

> The result (overflow cannot occur) holds true also with small computational delays.

Yes, I agree with you.

BTW I am still trying to get reliable send results with this library.  Still no joy.

bengtmartensson:
Checked in a new version. Among other changes, there is a spin wait option (that @MrBryonMiller  seemed to favor, see [Wikipedia](https://en.wikipedia.org/wiki/Busy_waiting) for some background), please try both with and without. Seams to produce fairly stable result, see screenshot, made by running IRsendDemo on a arduino.org M0 pro. . Note that the "marks" are much more precise than the spaces, which I believe is because  computational delays end up there (mostly).


screenshot from 2017-04-16 22-19-34

irSend.cpp
```diff
@@ -15,14 +16,57 @@ void  IRsend::sendRaw (const unsigned int buf[],  unsigned int len,  unsigned in
...
 }
 
+#ifdef USE_SOFT_CARRIER
+void inline IRsend::sleepMicros(unsigned long us)
+{
+#ifdef USE_SPIN_WAIT
+	sleepUntilMicros(micros() + us);
+#else
+	if (us > 0U) // Is this necessary? (Official docu https://www.arduino.cc/en/Reference/DelayMicroseconds does not tell.)
+		delayMicroseconds((unsigned int) us);
```
z3t0: Any reason to have both methods?

boarddefs.h
z3t0: I see your point. It's just that theres no proper way to deal with this in the Arduino environment.

> OK, we are all using something that is based on gcc, so you may argue that the argument is not valid...)

Right we might as well take that for granted

And better #warn then #error which will stop compilation? 

The reason i think its a good idea is that we can "hopefully" expect users to read the compiler output, and if necessary we simply have to point them to read it in new issues where the issue is inaccurate timings caused by soft pwm

z3t0:
FYI: for any new ideas that are offtopic I am documenting them at https://github.com/z3t0/Arduino-IRremote/projects/1 for future reference

irSend.cpp
bengtmartensson: The advantages and disadvantages of spin wait are described in in the Wikipedia article linked. @MrBryonMiller originally suggested using spin wait (without using that name). I think that the issue of using spin wait is bound to, "periodically", come up as soon as someone distrusts the system's delay*-functions. The way it is implemented is clean, and the user can just define USE_SPIN_WAIT to invoke it.

Summing up, "yes".

boarddefs.h
bengtmartensson: The technical arguments have already been listed. I am still against it. Let me give an example: In  building IrScrutinizer on a Linux machine, the wine program is invoked. (I use the Fedors 25 system version) It produces the following warnings: 

    fixme:winediag:start_process Wine Staging 2.4 is a testing version containing experimental patches.
    fixme:winediag:start_process Please mention your exact version when filing bug reports on winehq.org.
    fixme:file:SetSearchPathMode (8001): stub
    fixme:process:SetProcessDEPPolicy (1): stub
    fixme:file:SetSearchPathMode (8001): stub
    fixme:process:SetProcessDEPPolicy (1): stub

Irrelevant warnings hitting the wrong person, and just a royal PITA. I would not even consider to "react to" those warnings.

irSend.cpp
z3t0: Would the solution here provide any meaningful advantage. http://stackoverflow.com/questions/7083482/how-to-prevent-gcc-from-optimizing-out-a-busy-wait-loop

bengtmartensson: No. We do not have the problem addressed in that thread ("How to prevent GCC from optimizing out a busy wait loop?"), since micros() is not a const function (in C++ lingo).

z3t0: Ah I see, thank you for he explanation

MrBryonMiller:
I am officially confused. 
Point 1 - When @ladyada and I originally proposed a SAM change we used a method of providing a Mark wherein we would see if the Mark time had passed and if not would pulse the output high for 1/2 of a period then low for 1/2 of a period then check again to see if the Mark time had elapsed.  (I think of this issue as inter-period Mark termination) I thought @bengtmartensson was pointing out that our method was guaranteed to have a Mark that was too long by up to one period. He checked in a version that would end either the high pulse or the low pulse at the correct Mark ending time. Now, this latest version appears to me to lose that guarantee.  It appears that neither the high pulse nor low pulse will terminate early so this method will perform just like the original.  Am I missing something?
Point 2 - While I was trying to get @bengtmartensson's checked in code to work with my test program I proposed a different way terminating both high pulse and the low pulse at the proper time.  I felt my method had two advantages 1) without needing to throw away a Mark, it properly handled them even if there timing worked out such that the the Mark crossed the time when micros() wrapped around. 2) it was slightly easier to understand how it was accomplishing this (at least to me).  My proposed method happened to sit and spin in a while loop - but the avoidance of delayMicroseconds() was not the major point, rather the manner in which the loop ending conditions were calculated was.  This manner was such that inter-period terminations were properly handled and micros() wrap around were properly handled.  It appears to me that when @bengtmartensson attempted to accommodate my proposal (via use of USE_SPIN_WAIT) we lost my loop termination calculations thus both of my perceived advantages.

Recommendation: I think we need to decide whether we really need to end Marks at the correct time without the possibility of extending the Mark time by up to one period.  Even though the original code worked in my test environment I believe it is a good idea to so this.  Given that, I think we should have only one way to do this.  I can't imagine many people in the future understanding this point to this level and wanting to deal with defining a term to change the method.  And if someone does understand to this level and wants a change they'll almost surely want to define their own Mark timing with their own custom code.

bengtmartensson:
@MrBryonMiller :

The problem with my original code was that when "time" had passed and the last part was OFF, the generated pulse was effectively too short, since the last silence was effectively concatenated with the following space. There is no way to distinguish between "mark-silence" and "space-silence", so effectively the generated marks were on average too short. Your code  however, in the presence of computational delays, generated too long single pulses, i.e., too low modulation frequency. @ladyada discovered this empirically, and added an ad-hoc correction, which is not a robust solution. My code (all versions) end the pulses at starttime + count*periodTime, which _is_ robust against computational delays. So, yes, you missed something. I think (now!) that generating the correct frequency is the highest priority. How to ideally represent a mark when the signal is off halt of the time is not a well defined problem. Just generating until micros() >= stop will, as motivated above, give too short signals on the average; therefore in the last version I wrote now < stop + periodOnTime / 2, to get it more right "on average". 

> I can't imagine many people in the future understanding this point to this level and wanting to deal with defining a term to change the method. 

My code encapsulates the use of different delay algorithms. 

> And if someone does understand to this level and wants a change they'll almost surely want to define their own Mark timing with their own custom code.

Either a low-level, hardware dependent, efficient code (hard-carrier), or a high-level, portable code (soft-carrier). This is the latter.

MrBryonMiller:
> The problem with my original code was that when "time" had passed and the last part was OFF, the generated pulse was effectively too short, since the last silence was effectively concatenated with the following space.

Isn't this exactly what happens when there is a HW PWM?

I want to be sure I understand this.  Your latest code - once it decides to do another iteration will stay in the loop for a time equivalent to a high pulse plus a low pulse? This is a period of time we referring to as a period. So it seems to me, if we check the loop re-entry condition just as mark time is about to expire, we will spend one whole extra period in the loop.  I think this is equivalent to saying we will be spending, on average, an extra 1/2 period inside of the loop. Right?  Since the loop termination is now "stop + periodOnTime / 2" aren't we going to be spending, on average, one whole period extra in the loop?

> Your code however, in the presence of computational delays, generated too long single pulses,

I don't really want to argue about this point, but just as a point of reference the computation delays I was seeing were less than a microsecond.  In other words the same order of magnitude we get by calculating microseconds using integer math.

> I think (now!) that generating the correct frequency is the highest priority. 

Again, I don't have an argument here, just wondering why is frequency more important that mark width? Isn't accurate frequency a matter of the bandpass on the IR receiving devices? Are they really that tight?

bengtmartensson:
> Isn't this exactly what happens when there is a HW PWM?

You may have a point here. 

> aren't we going to be spending, on average, one whole period extra in the loop?

we are spending more time in the loop, yes. But feel free to improve my code, which is to be seen as a suggestion. 

> but just as a point of reference the computation delays I was seeing were less than a microsecond. 

Changing the period time from 25 to 26 micros changes the corresponding frequency from 40 to 38.461 kHz, which is not a small change.

> In other words the same order of magnitude we get by calculating microseconds using integer math.

Talking about periodTime = 1000/khz, right?  Difference being that this is a completely deterministic,  hardware independent, quantization. It does not change if the CPU frequency is changed.

Have you tested the new code?

MrBryonMiller:
> Have you tested the new code?
I converted my test setup to do a little PWM work.  I will convert it back and test new code but it won't be for a while

MrBryonMiller:
> Have you tested the new code?

Yes - my test setup sends a single NEC code 50 times in a row  from a Feather M0.  Receiver is an Arduino Uno only inches away. I tested the latest configuration and I'm including the results from an earlier test using the library built by @ladyada.  I am including this just as confirmation that my test setup is not completely out of whack.

With the ladyada library I got 49 correct out of 50.  Latest configuration of this library 36 correct out of 50.

I'm trying to examine the raw data to see if I can tell what is making the failures occur.  I assume there is something off about the Mark timing but maybe it has more to do with the Space timing.  I am including an IRrecvDump of a reception that worked and one that failed.  Maybe someone can spot what might have caused the failure for this particular case.

```
IRrecvDumpV2 listening
Encoding  : UNKNOWN
Code      : C8F63E54 (32 bits)
Timing[67]:
     +9000, -4400     + 650, - 450     + 550, -1700     + 650, -1600
     + 550, - 550     + 550, - 600     + 600, - 500     + 650, - 500
     + 650, -1600     + 550, -1700     + 550, - 550     + 550, -1700
     + 550, - 600     + 650, - 450     + 650, - 500     + 650, - 450
     + 500, - 650     + 650, - 450     + 500, - 650     + 600, -1600
     + 550, - 600     + 550, - 600     + 500, - 600     + 500, - 650
     + 650, - 450     + 550, -1700     + 650, -1600     + 650, - 450
     + 650, -1600     + 550, -1700     + 550, -1700     + 650, -1600
     + 650, -1600     + 550
unsigned int  rawData[67] = {9000,4400, 650,450, 550,1700, 650,1600, 550,550, 550,600, 600,500, 650,500, 650,1600, 550,1700, 550,550, 550,1700, 550,600, 650,450, 650,500, 650,450, 500,650, 650,450, 500,650, 600,1600, 550,600, 550,600, 500,600, 500,650, 650,450, 550,1700, 650,1600, 650,450, 650,1600, 550,1700, 550,1700, 650,1600, 650,1600, 550};  // UNKNOWN C8F63E54


Encoding  : NEC
Code      : 61A020DF (32 bits)
Timing[67]:
     +9000, -4350     + 650, - 500     + 550, -1700     + 650, -1550
     + 700, - 450     + 500, - 600     + 650, - 500     + 550, - 550
     + 650, -1600     + 650, -1600     + 650, - 500     + 650, -1600
     + 550, - 550     + 550, - 600     + 650, - 450     + 650, - 500
     + 650, - 450     + 650, - 500     + 650, - 450     + 550, -1700
     + 650, - 500     + 600, - 500     + 650, - 500     + 600, - 500
     + 650, - 500     + 650, -1550     + 550, -1700     + 550, - 600
     + 650, -1600     + 650, -1600     + 500, -1700     + 600, -1650
     + 650, -1600     + 650
unsigned int  rawData[67] = {9000,4350, 650,500, 550,1700, 650,1550, 700,450, 500,600, 650,500, 550,550, 650,1600, 650,1600, 650,500, 650,1600, 550,550, 550,600, 650,450, 650,500, 650,450, 650,500, 650,450, 550,1700, 650,500, 600,500, 650,500, 600,500, 650,500, 650,1550, 550,1700, 550,600, 650,1600, 650,1600, 500,1700, 600,1650, 650,1600, 650};  // NEC 61A020DF
unsigned int  data = 0x61A020DF;
```

bengtmartensson:
Just a quick remark: Remarkable are the large "random" deviations, and the short spaces. Cf. my screenshot above, using sony12 using IRSendDemo; as you can see the times are much more stable. But I am almost certain that it is the 450 spaces that makes the decoder reject the signal. I have right now no explanation for this -- we have not been fumbling with the space durations at all! "This cannot happen"!! 

What happens if you remove the "+ periodOnTime / 2" in the comparision?

MrBryonMiller:
> What happens if you remove the "+ periodOnTime / 2" in the comparision?

```
     waiting for           |   waiting for  |    waiting for 
stop - periodOnTime / 2    |      stop      | stop + periodOnTime / 2
     36/50                 |      32/50     |      36/50 

```
in other words all basically the same.  So like you say, it seems like the space timing which I agree has not been altered.

Neither do I get this.  Also confusing, if I replace the entire Mark code with the either the ladyada code or the code I proposed earlier the results improve to at least 49/50.

> it is the 450 spaces that makes the decoder reject the signal

One more confusing thing is that many of the receptions that work (including the one I captured above) also had a few 450's.  Do short spaces accumulate?

I hope this is not something funny in my setup - can you transmit and Scrutinize a NEC code, maybe even the same one I used (0x61A020DF)?

AnalysIR:
Depending on how much current you are putting thru the IR LED, "inches away" could be part of the problem.

I usually advise a minimum of 1m/3ft for tests. (Overloading of the IR receiver internals can occur close up)

Also, the times you are getting are well within the spec of most IR receivers - so that would point a finger at the decoder tolerance.

FYI: On AnalysIR both of these signals decode as NEC 0x61A020DF.

bengtmartensson:
> can you transmit and Scrutinize a NEC code, maybe even the same one I used (0x61A020DF)?

Yes. Using my M0 Pro, code as the present git,  and the setup previously described (important Options -> invoke signal cleaner is off).

screenshot from 2017-04-18 11-13-58

as you can see, very stable values, close to their theoretical values. What happens is we use a demodulating decoder instead (IrScrutinizer: select "Use receive for capture"). This happens:

screenshot from 2017-04-18 11-19-24
 
As you can see, lesser precision in the measurements, but still stable, and no 450. (The frequency value of 38000 is a dummy value.)

> FYI: On AnalysIR both of these signals decode as NEC 0x61A020DF.

And so does DecodeIR/IrScrutinizer. (Or, rather, the correct parametrization, device = 134, subdevice = 5, commandnumber = 4.)

> I usually advise a minimum of 1m/3ft for tests. (Overloading of the IR receiver internals can occur close up)

You can try just holding a sheet of paper in front of the sensor.

MrBryonMiller:
@AnalysIR, @bengtmartensson, I was driving the LED from 3.3 volt output through 330 ohm resistor, assuming about 2v forward drop on IRLED, I figure about 3ma.  Changed to different RX device, changed R to 1k, changed distance to max convenient of 8 inches,  put a piece of paper between LED and RX.  All results seemed the same mix of good decodes and bad decodes although I did not quantify results.

ARRRGH - What is going on??

AnalysIR:
More like 6mA with the 330R. Either way very little current. I would suggest driving it thru a transistor to recreate 'real-world' comparisons. (6mA is probably not a valid test scenario)

As before, the times you are getting are not really out of spec, and are 'failing' because of the tolerance of IRremote, which you should be able to adust/increase to suit.

try increasing mark excess from 100 to **150**

`#define MARK_EXCESS 150  //in IRremoteInt.h`

_**Note**: 560-450 = 110, so 150 should be safe enough for your setup!_

MrBryonMiller:
@AnalysIR 

> try increasing mark excess from 100 to 150

100% **errors** - although maybe 10% of the received values said the received value was JVC format with a value of 0x61A0 in other words the first 16 bits of expected result.

AnalysIR:
OK, thats a pity. I would then suggest using a Scope or LA to examine the Tx & Rx waveforms in detail, which should show what is happening.

What model of IR receiver are you using? (Link?)

MrBryonMiller:
@AnalysIR I think you are right - next step is a `scope.

IR receiver is CHQ 1838 - all links to data sheets I find are in Chinese.  Maybe you can familiar with device.

bengtmartensson:
> IR receiver is CHQ 1838 - all links to data sheets I find are in Chinese. Maybe you can familiar with device

I gather up for the storm... :-)

MrBryonMiller:
@bengtmartensson Before retiring to my lab I was taking another look at your latest change.  I noticed
```
	if (us > 0U) // Is this necessary? (Official docu https://www.arduino.cc/en/Reference/DelayMicroseconds does not tell.)
		delayMicroseconds((unsigned int) us);
```
Since us is unsigned won't it always be >= 0? So test is unnecessary.  Or are you worried about ==0?

> I gather up for the storm... :-)

I take it this device has a reputation?  Maybe so but I have been getting excellent results with ladyada's code.

bengtmartensson:
> Since us is unsigned won't it always be >= 0? So test is unnecessary. Or are you worried about ==0?

Yes, it is the test for zeroness that is relevant. If you test for >= on an unsigned, the compiler will detect it and warn, and not generate code for the always true test.

The remark complains about sloppy documentation. It is really *undocumented* what the function does with a zero argument! Unfortunately, the Arduino documentation is full of similar examples. Is this what they consider "aimed at the beginner"??

If you study earlier issues, you will discover that @AnalysIR several times has commented on the quality of cheap no-name receivers, and recommends a certain Israelian brand (that is not a criticism!)

AnalysIR:
:)  
Yes it has attracted a range of opinions here. Mine is that it is among the worst and results in lots of wasted time on support areas like this.
It does work perfectly in some situations (preferably very dark rooms), and it is extremely cheap for a reason.

IMHO, there is no point troubleshooting based on this device. TSOP34438 would be a good alternative. (circa US$0.50)

I have ocassionaly noticed over the years that minor glitches/blips in carrier waveform generation can cause issues during demodulation, even with quality receivers. Perhaps the differing behaviours you are seeing are related to something similar??. 

FYI: [https://www.analysir.com/blog/2014/12/08/infrared-receiver-showdown-tsop34438-vs-vs1838b-winner-revealed/](https://www.analysir.com/blog/2014/12/08/infrared-receiver-showdown-tsop34438-vs-vs1838b-winner-revealed/)

PS: I always assumed Vishay was a derivative of Telefunken (at least the IR stuff). Either way they do seem to be the quality/primary supplier for IR Receivers.

MrBryonMiller:
> IMHO, there is no point troubleshooting based on this device.

I have no problem taking my problems off line.  But since the whole point of this issue is the acceptability of  @bengtmartensson's PR doesn't someone have to hook an Arduino-IRremote receiver to an Arduino-IRRemote sender (where the sender is a SAM machine) to make sure things work reliably?  Or is it sufficient to use analyzers and/or scrutinizes to declare it good?  They make for a powerful argument but I always like the warm-fuzzy feeling seeing it work.

And now that I think about it, given that the interrupt structure is entirely different from the machines the Library previously supported, someone needs to take a reasonably detailed look at reception on a SAM machine (The bulk of the testing I have been doing is sending from a M0 while receiving on an UNO).  I have received using @bengtmartensson's library on my M0 but only a few buttons from a Remote and all using my low-quality IR receiver device.

> I have ocassionaly noticed over the years that minor glitches/blips in carrier waveform generation can cause issues during demodulation, even with quality receivers.

Yes, the minor glitches/blips is what I am worried about.  So I am going to continue my efforts with a oscilloscope to see if I can see anything hinky. Thanks to the @AnalysIR's point I'll look at the output of one of my cheap receivers while I'm studying the send waveform.  If I find anything untoward in the send waveform I'll bring it up here.  Assuming it looks good but I see something wrong with the received waveform I'll place a Mouser order.

bengtmartensson:
> Or is it sufficient to use analyzers and/or scrutinizes to declare it good? 

The answer is "yes". I have tested with an M0 Pro (arduino.org), which produced reliable result. When it comes to analyzing an IR signal, IRremote's irRecv is one of the bluntest tool around, also with a sane IR receiver. (Anything with a de-modulating receiver is "blunt".)

> I'll place a Mouser order.

Be sure to order a few TSMP58000. (See e.g. [this guide](http://www.harctoolbox.org/arduino_nano.html).)

MrBryonMiller:
@bengtmartensson I did not realize the default for the soft carrier was to use spin wait.  I did not realize this until I saw your last commit.  I figured that might have been contributing to me getting inconsistent results (in addition to the cheap receiver devices).  So I tested your latest with my program that sends 50 NEC values in a row.  What I found was that anywhere from 1 to 4 times out of the 50 the program would get hung up in the send routine.  I could wait it out but each time it took right around 90 seconds before it would continue on.  I instrumented enough to watch the value of micros() when the hangup occurred and micros was nowhere near wrapping around its 32 bit value. (as one would expect since the program only runs a few minutes, nowhere near 4,000 seconds.)  I've studied your mark code in some detail and can't see anything might cause this.  I thought that maybe there was something in you library that cause the receive interrupt to continue to run during send but could not find anything wrong with that either.

Have you seen anything like this? Anyone else?

bengtmartensson:
Just a few quick comments:

* Please show us your sketch
* Please try both with and without spin wait,

> something in you library that cause the receive interrupt to continue to run during send
?? Are running receive simultaneously to sending?

MrBryonMiller:
> Please show us your sketch

```
#include <IRremote.h>

#define LED_PIN 13

int RECV_PIN = 11;         

IRrecv My_Receiver(RECV_PIN);
IRsend My_Sender;

void setup() 
{
// start serial port
pinMode(LED_PIN, OUTPUT);
Serial.begin(9600);
int waitcnt = 0;
while(!Serial && (waitcnt++ < 50))  // wait (only so long) for serial port to connect.
     delay(100);
Serial.println("Serial Started");

for (int i=0;i<50;i++)
     {
     Serial.print(i);
     Serial.print(" ");
     unsigned long s = micros();
     sendcode(0x61A020DF);
     Serial.println(micros()-s);
     delay(250);
     }

My_Receiver.enableIRIn(); // Start the receiver
My_Receiver.blink13(0);
}

void loop() 
{
}

void sendcode(long code)
{
Serial.print("Sending ");
Serial.println(code,HEX);
My_Sender.sendNEC(code,32); 
Serial.print("delay ");
Serial.print(micros(),HEX);
delay(500);
Serial.println(" Sent");
}

```

> Please try both with and without spin wait,

Works as expected when USE_SPIN_WAIT is defined.  Hangs up randomly when USE_SPIN_WAIT is commented out.

Please note, when I say works as expected I mean sending program progresses normally.  But there are still a high level of receive errors.  TSOP4838 should be here tomorrow.

> Are running receive simultaneously to sending?

No, but it'd be nice if we could.

bengtmartensson:
@MrBryonMiller :
I see several problems. You are mixing different time sensitive operations; sending as well as Serial prints. Using 9600 bits/s may also be a problem. There is a lot of timing involved in Serial, involving timers... Please delete all console output and try with just

    void loop() {
        irSend My_Sender.sendNEC(..., 32):
        delay(....);
    }

Please also remove the irRecv instance. To be on the safe side, no not instantiate Serial at all.

Just as @AnalysIR  said that he will not accept results from a cheapie IR receiver, I will not accept results with intermingled console output. 

> Works as expected when USE_SPIN_WAIT is defined. Hangs up randomly when USE_SPIN_WAIT is commented out.

Which at least shows that it can be valuable to have it, at least as a troubleshooting option.

MrBryonMiller:
@bengtmartensson 

Following still gets hung up every 3-10 times through loop.

```
#include <IRremote.h>

IRsend My_Sender;

void setup() 
{
for (int i=0;i<50;i++)
     {
     My_Sender.sendNEC(0x61A020DF,32); 
     delay(500);
     }
}

void loop() 
{
}

```

bengtmartensson:
@MrBryonMiller 
sorry, I am hardly able to help you at this point. Possibly you are running out of stack or something like that. What exactly does "gets hung up" mean? 

Again, it runs fine for me using, I think, the same processor. I received a M0 Feather, but I have not have time to test it.  After all, this is about the PR, not sorting out your problems... 

How do we proceed with the PR?

MrBryonMiller:
@bengtmartensson 

I make sure I did not contaminate your library while I was making comment changes to SPIN_WAIT I downloaded your latest commit and reran my tests.

> Possibly you are running out of stack or something like that. 

Perhaps - but if I change the IR library to @ladyada's library there is no anomalous behavior.

> What exactly does "gets hung up" mean?

It is impossible to tell with the version of the sketch you asked me to build that has no Serial.prints other than occasionally the receiver shows nothing being received for 90 seconds or so.  For the version of the program that has prints, for those iteration where a hangup occurs, I can see a print is made just before calling sendNEC and about 90 seconds after the return from sendNEC. The prints after the call shows a little more than 90 million microseconds has elapsed.  Please remember this 90 second problem does not occur on every send.  At least 40 out of 50 seem to work fine (although many with receive decode errors).

> Again, it runs fine for me using, I think, the same processor.

I also think we are using the same processor - but if you install the Feather M0 following AdaFruit's instructions you do get some different core libraries. I **_think_** I looked at the implementations of micros() and delayMicroseconds() for the two cores and I could not tell a difference, but what do I know.

> How do we proceed with the PR?

I am going to rerun my test using using a TSOP34438 on my receiver end and using SPIN_WAIT from the @bengtmartensson library to see if the reception errors clear up (I should have results shortly).  Depending on the results of that I would ask deferring a decision until someone can confirm the library does not conflict with something on a Feather M0.  If an immediate decision does need to be made and if you wanted SAM support and you felt that support should include a Feather M0 and if it were up to me I would have to recommend @ladyada's PR over than this one. I think that would be a shame since we should all recognize that @bengtmartensson did an awful lot of cleanup and generalization work.

MrBryonMiller:
I have retested with TSOP4838's.  I used @bengtmartensson's library with one edit - I defined USE_SPIN_WAIT.

50 out of 50 were received correctly.  Thank you @AnalysIR!

> How do we proceed with the PR?

I would change my recommendation - If an immediate decision does need to be made and if you wanted SAM support and you felt that support should include a Feather M0 and if it were up to me I would have to recommend either this PR with change of using SPIN_WAIT being the default or @ladyada's PR.  I would be willing to reconsider the change of my USE_SPIN_WAIT recommendation if someone can confirm that the Feather M0 does not get hung up.

MrBryonMiller:
@bengtmartensson 
I found what is causing the hangups when not doing SPIN_WAIT.  In your loop you calculate nextPeriodEnding.  On occasion (usually less than 1 time out of 10), during the last time through the loop, this value can actually be less than the current value of micros(). When that happens, your calculation of targetTime-micros() using unsigned math ends up being a very large number.  Hence the very long delay.  (I have no explanation why you did not see the hangups other than possibly your M0 is somewhat faster so slower than mine so you never get into the case I run into.)

I fixed this by changing your test in sleepMicros from `if (us > 0U)` to `if ((us > 0U) && (us < 0x0000ffff))`
When I made this change my 50-in-a-row test ran every time I tried it without hanging up and thanks to the higher quality sensor on my receiver, No errors!

But what I really wish you would consider is replacing your entire loop, including the USE_SPIN_WAIT distinction, with the code I suggested earlier.  I have re-included it here with some added comments and updating the 'now' variable name and using SENDPIN ON and OFF to be consistent with your work.

```
long now = micros();
long markStop = now + (long) time;
// send out HIGHs and LOWs for the entire time of the Mark
// note the use of signed math so that wrap-arounds, when they occur, are properly handled
while ((markStop - now) > 0) { 
	SENDPIN_ON(sendPin);
	long HIGHStop = now + periodOnTime;
        // send out the HIGH for one half the period time OR until the Mark time is exceeded
	while (((HIGHStop - now) > 0) && ((markStop - now) > 0))
		now = micros();
	SENDPIN_OFF(sendPin);
        // send out LOW for one half period time OR until the Mark time is exceeded
	long LOWStop = now + periodTime-periodOnTime;
	while (((LOWStop - now) > 0) && ((markStop - now) > 0))
		now = micros();
	}

```

I know it has open loop timing considerations, but I feel they are small. And it has the advantage of terminating the Mark when its time is up, even if its in the middle of HIGH time - just like a HW PWM implementation would.

bengtmartensson:
> On occasion (usually less than 1 time out of 10), during the last time through the loop, this value can actually be less than the current value of micros(). When that happens, your calculation of targetTime-micros() using unsigned math ends up being a very large number. Hence the very long delay. 

_EXCELLENT!! Explains_ what happens without any guesswork, which is very satisfying, also theoretically. (Just slightly embarrassing for me...:-( ). I selected another fix though.

> But what I really wish you would consider is replacing your entire loop, including the USE_SPIN_WAIT distinction, with the code I suggested earlier. 

Actually, I consider my code better. It encapsulates the waits better, allowing for a clean selection between spin wait and normal wait. Also,

> long now = micros();

is outright wrong.

MrBryonMiller:
> EXCELLENT!! Explains what happens without any guesswork

Yes - mystery is still where did 90 or so seconds come from.  Looking inside of delayMicroseconds() I 
 get a glimmer of an answer due to 2^32 / 48000000 = 89. Where 2^32 is the approximate value (actually 2^32 minus a couple) that is argument to delayMicroseconds when the error occurs. And 48000000 is the clock speed of the M0.  Still don't pretend to understand exactly.

> Actually, I consider my code better.

I respectfully disagree - but I give up.

> long now = micros(); is outright wrong.

I might not have done this in the "right" way but I needed to perform signed math so wrap-arounds worked out correctly.  So I needed now to be a long.  But again, I give up.

Bottom Line : I looked at @bengtmartensson latest commit - I have tested send and receive with a Feather M0.  I have previously done some level of regression testing with other machines.  I believe this commit would make a fine addition to this great Library.

bengtmartensson:
Just a final remark:

> I might not have done this in the "right" way but I needed to perform signed math so wrap-arounds worked out correctly. So I needed now to be a long.

AFAIK, what happens when a signed variable in C (C++) "wraps around" is undefined. There might even be hardware "traps" fired when wrapping around. C is not Java, which is guaranteed to use 2-complement. There might be compiler flags that, for a certain processor, forces a particular behavior. But, if the aim is to write portable code (that runs on all processors) -- which should be our aim -- stay out of undefined behavior.

bengtmartensson:
Any progress on reviewing/merging this?

z3t0:
@bengtmartensson Sorry I am currently busy with things in life and will not have a chance to merge or review any times soon.

I would suggest we go ahead and create a beta release with the changes in this merge and add that to the releases.

Thoughts?

bengtmartensson:
Almost 3 years has passed, and no sign of a merge... I have implemented all the features: SAM support send and receive, soft carrier PWM, and much more, in the just released version 1.1.0 of my  library [Infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino). (Available in the Arduino library manager in a few hours).

PR #481, opened 2017-06-22
Add receiving support for Arduino 101

jimaobian:
Add receiving support for Arduino 101
Add F() in the serial print
Refine the macro define for esp32

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -24,7 +24,7 @@
 #	include "IRremoteInt.h"
 #undef IR_GLOBAL
 
-#ifndef IR_TIMER_USE_ESP32
+#if (!defined(IR_TIMER_USE_ESP32) && !defined(IR_TIMER_USE_CURIE))
 #include <avr/interrupt.h>
 #endif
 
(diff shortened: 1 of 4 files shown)


z3t0:
There has not been a formal review process setup for merging new additions to this project, but I think it would also be great if we could some confirmation that this also works for other users. Though again, I have no idea who from the community has access to the hardware (Arduino 101)

@jimaobian  thoughts?

bengtmartensson:
This PR is not compatible with PR #437. Instead, it adds yet another C pre-processor symbol to the code proper (`IR_TIMER_USE_CURIE`) which was exactly the problem that #437 tried to address. 

#437 adds a "soft-carrier-send" which should make sending possible on all sufficiently fast boards, like the Curie.

jimaobian:
@z3t0  Yes, I agree. This PR should be a hot patch for someone who would like to use IRRemote library on Arduino 101. It should be never merged but still exist until there are better implements.

@bengtmartensson Great! It is more elegant.  I'll do some review after #437 is merged into master.

z3t0:
@jimaobian added to the readme

jeffjvick:
Hi, I tested this 'hot patch' on an Arduino 101 and the receive appears to be working.  Unfortunately what I really need is to send.  

@jimaobian have you had any chance to look at what changes are needed on top of #437 to support this?  Is there any quick dirty way to send some NEC codes out?  I just need something hacky to get by.  Thanks.

z3t0:
I created a beta release for #437 

If it is okay I would like to close this PR as a soft carrier has already been implemented in said PR.

yygcom:
I need send , Oh my god

bengtmartensson:
Suggest closing this PR. It is not compatible with the current code in master.

Issue #464, opened 2017-04-24
Make source file layout compliant with Arduino library standard


bengtmartensson:
There _is_ actually an [Arduino library specification](https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification). According to that, the sources should be a in a subdirectory names `src`, and the non-public (not exported) header files in sub-directory/sub-directories of that directory. 

This should possibly replace #375 (which I am personally find very unsatisfying).

Not really urgent.

z3t0:
Is there an official reference document somewhere to follow, or should I just conform to the layout of official libraries

z3t0:
This might be what we are looking for: https://github.com/arduino/Arduino/wiki/Arduino-IDE-1.5:-Library-specification

bengtmartensson:
That was the document I linked to in the original post ;-).

z3t0:
Ah just saw that... oddly the original post was blank when I originally saw it, just the title was visible

z3t0:
Please review the commit 635e8a9 and then we can close this issue.

bengtmartensson:
The most important thing I believe is not the `src` directory but the hiding of the "private" includes (`boarddefs.h, IRremoteInt.h`) in a sub directory. I just read the library guidelines again, and I see that there is now an alternative method. (I cannot recall that it was there when I wrote the OP.) 

> includes - (available from IDE 1.6.10) a comma separated list of files to be added to the sketch as #include <...> lines. This property is used with the "Include library" command in the IDE. If the property is undefined all the headers files (.h) on the root source folder are included.

So I ask you to implement one of these methods. I think the sub directories is the preferred one: It works for IDE < 1.6.10, and it is more obvious to the user that some includes are public, some not. The drawback is that all the corresponding include statements need to be changed.

z3t0:
My preference is for the first method where we use subdirectories to hide
the libraries own headers, that way the top level headers are much more
accessible and there is a clear distinction in the software design.

Thanks,
Rafi Khan


On Sun, Aug 13, 2017 at 11:56 AM, Bengt Martensson <notifications@github.com

bengtmartensson:
Has been fixed by PR #507. I take the liberty to close.

PR #507, opened 2017-08-18
Moving private includes to sub-directory,  #464

bengtmartensson:
This PM implements the sub-directories as discussed in the context of #464.


bengtmartensson:
> Not much point in having paragraph if it's just a duplicate of sentence?

The reason that I added it was that the Arduino IDE (1.8.3) complained. (I started arduino from an xterm, if you start it by double clicking or something, it may happen that these warnings go into nirvana.) If "paragraph" is optional, then the IDE behaves incorrectly.

The best thing would probably be to add a "paragraph" (= several sentences) stating something more non-trivial.  Personally, I do not really care.

z3t0:
That's odd, what about this then as the paragraph:

Arduino-IRremote has built in support for sending and decoding many IR signals.

bengtmartensson:
I just revoked the change to library.properties, in order to keep the 1-1 correspondence to #464. (It was just a ugly warning that I wanted to kill.) After all, the warning does not hamper proper operation.

Still think that it should be fixed, though.

z3t0:
> Still think that it should be fixed, though.

Could you open an issue with the warning output? That way we can explore this while also merging these changes.

PR #512, opened 2017-08-22
Doxygen API documentation in the sense of #461

### LINKED ISSUE #461: API documentation using Doxygen
Writing API documentation is a major TODO, as I see it. I would here like to recommend [Doxygen](http://www.doxygen.org), which generates the API documentation from Javadoc like comments. For an example, see [Infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino), with the produced documentation available [here](http://harctoolbox.org/infrared-apidocs/index.html). The documentation is generated at the developer's PC. and (for example) uploaded. The above mentioned Infrared4Arduino contains the Doxygen runs in the Makefiles.

I can contribute the main configuration, the `Doxyfile`. Documentation for the individual classes/function do not need to be written at one time, also withoth the comment the output is usable.

Writing a "book" on the API never works: either it will never be finished, or is obsolete by the time on finishing... (See IRLib for an illustration.)  Generated API docs is the way to go (agile).

bengtmartensson:
This PR contains a usable Doxyfile and changes to two important files in order to document the API.

Should fix #461, although much documentation remains to be written.


z3t0:
Thanks for taking the initiative!

PR #516, opened 2017-08-27
boarddefs.h cleaning and documentation

### LINKED ISSUE #444: Reorganizaton of boarddefs.h
Sort-of continuation of #437.

It turns out that the stuff that must be added for supporting a new board is sprinkled all through the file. Probably this is due to the time when all processors where AVR, and most configurations were dependent on the timer selected. 

To address this problem, I would like to suggest a reorganization like:

    // Default definitions that only a few boards like to change

    #if defined(BOARD1)
    // stuff for BOARD1
     #elif defined(BOARD2)
    // ...
    ....
    #else
    #error No board recognized
    #endif 

    // Fallback definitions, like
    #if defined(SEND_PIN) & !defined(SENDLED_ON)
    #define SENDLED_ON(pin) digitalWrite(SEND_PIN, HIGH)
    #endif 

alternative structure: Everything that concerns a certain board is in a certain file, i.e.,

     #if defined(BOARD1)
     #include "board1.h"
     #elif defined(BOARD2)
     #include "board2.h"
    ....
    #else
    #error No board recognized
    #endif 

    ....

 Feedback wanted.

bengtmartensson:
This PR contains cleanups and documentation (in the sense of #461) of boarddefs.h and IRremoteInt.h.
It is *almost* code neutral, the exceptions being:

1. If ARDUINO is *not* defined, it is assumed that we are compiling a test version for execution on the host; timers and blinking parameters being nullified,
2. Previously, if no sensible pre processor symbol was found, a AtMega382p-like board (e.g. Uno) was assumed. I do not thing this is a sane decision any more. Instead, the __AVR_ATmega328P_  and __AVR_ATmega168_ _ are explicitly treated, and the "default" case triggers an `#error`.  (However, for blinking, LED_BUILTIN is assumed (if defined).)

Should be considered to resolve #444.


z3t0:
Will review this in a few days once I get a chance as this is has quite a lot of changes.

ArminJo:
Hi,
can you redirect the base of this PR from beta to dev, since the beta branch is going to be deleted?
Thanks.

PR #522, opened 2017-09-14
Added support for RC5 extended

LarsWH:
Hi,

I added support for 'RC5 extended' in order to make this project: https://www.instructables.com/id/Remote-Control-for-Lava-mMotion-Swing-Mounting-Bra/

The modification is working nicely in this single application. It has not been tested elsewhere.

(I am new to Git and this is my first pull request ever. So please excuse me if I have messed something up)

Kind regards,
Lars

PR #527, opened 2017-09-24
Merge #522

z3t0:
See #522


z3t0:
Something definitely seems to be broken... Out of time for now but will take a look later.

z3t0:
the dev branch is old... using beta now I think

ArminJo:
I compared it with the master head and all changes seem to be included now. 👍

PR #533, opened 2017-10-09
Add RCA protocol (for e.g. Thomson TVs)

grensjo:
The protocol used by my TV remote was not supported by this library. After some googling I ended up at these descriptions of the RCA protocol:
* http://www.sbprojects.net/knowledge/ir/rca.php
* http://www.numericana.com/answer/ir.htm#rca

The protocol is very similar to NEC, so I started from `ir_NEC.cpp` and modified it to work with RCA. This implementation is verified to work with my Thomson TV (model FZ3234, remote RC3000E02). The author of the first link above mentioned that (some?) Xbox remotes also use RCA, but I have no other confirmation of that.

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -79,6 +79,9 @@
 #define DECODE_LEGO_PF       0 // NOT WRITTEN
 #define SEND_LEGO_PF         1
 
+#define DECODE_RCA           1
+#define SEND_RCA             1
+
 //------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ArminJo:
If you rebase your code, I can merge it.
And do not forget to modify Readme, changelog and keywords.txt

ArminJo:
Still intersted in this PR?

PR #540, opened 2017-11-27
ESP32 Support

SensorsIot:
I added ESP32 support (using ledC PWM). You can now select any pin in IRsend ().
Maybe you want to watch it in use: https://youtu.be/gADIb1Xw8PE


wybiral:
Any update on this?

requeijaum:
Will test it today in some project of mine.

1uuc4asb:
I'm in @requeijaum 's project .
I tested and it worked well!
I'll make a video soon.

toomasz:
Please also update readme.

Sam-espressif:
Will it be implemented using RMT peripheral ?

lewisxhe:
Great, I have tested it to control the air conditioner.

AnalysIR:
@makerMcl looks like the duty cycle is set to 83% (not 33%) ???


> patch for IR send duty cycle to 33% 

`ledcWrite(LEDCHANNEL, 83);`

makerMcl:
@AnalysIR: sorry, the argument is no percent. The base for this value is set with the 3rd argument in `ledcSetup(LEDCHANNEL, khz*1000, 8);` in irSend.cpp, so it's 83/255=32.5%.
Originally I found the problem source with a logic analyzer, it confirms the values above. You can find these screenshots at https://github.com/makerMcl/Arduino-IRremote/tree/master/test

AnalysIR:
OK, thanks for the calrification :)

AnalysIR:
...just a thought

Is it integer division?   If so you will end up with 32%.

using a value of 85...should get you 33.33% exactly (even with/without integer division)

makerMcl:
Yeah, that is right. As I understand the ESP32 spec, it will be no division but might be implemented with a counter register - the resolution parameter of 8 (bits) sets the counter width and duty is determined by comparison. In fact, I guess that is theoretical, because:
RC5 documentation on wikipedia says it's a duty cycle of 25%, and that is matched with the original value of @SensorsIot.
In reflection I have to admit I used a simple transistor driving circuit, without the bypass-diode between base and collector. That results in less-steep flanks and diminishes the duty cycle. You can see the effect on the pictures contained in https://github.com/makerMcl/mmControl/tree/master/docs.
So if I want it to be perfectly done, it should be parameterized according to the circuitry between GPIO pin and ir diode, e.g. distinguishing between direct/slow flank/steep flank.
I guess I will check it soon again (the original value of SensorsIot ) with an implemented circuit and prepare another pull request for that parameter...

AnalysIR:
yes..25% is correct based on Philips docs. However, the original stated motivation was to **minimise** power usage, which this does. Other philips docs also mention both 50% & 25% for RC5.

However, **for low power** (battery) usage - 33.3333% is the **optimum** duty cycle and the emitter circuit should be designed based on that where possible.

If **not** desiging for low power/battery usage then blasting away @ 50% is my preferred choice.

The main difference for the IR receiver between different duty cycles is the relative amount of IR being received @ the nominal frequency (vs power wasted in harmonics from the square wave).

**_Guessing_**: _early on they were using specific oscillators which allowed them to get ~36kHz carrier. The IC then only allowed them to get 50% or 25% (assuming no available integer divisor for 33.3%). So in this case they chose 25% as lower power than 50%.....just a guess.  MCUs are more capable these days, so these types of restrictions are less relevant/common._

makerMcl:
Fine, that's one more point for a general paramter to control the duty cycle in the lib. I have not seen that in the base implementation, did I miss it? (I am sorry I can contribute the complete implementation only later, so anybody feel free to add and I will adapt it on the Esp32 implementation)

makerMcl:
Hello,
I made some integration tests and found a timing problem - at a sequence of 3 ir commands, the first command is crumbled by a oversized mark or space. The difference is about 5 microseconds.
I made a workaround, you find it at https://github.com/makerMcl/Arduino-IRremote/tree/esp32-timing-problem. Screenshots of the problem are at https://github.com/makerMcl/Arduino-IRremote/tree/esp32-timing-problem/test/esp32-timing-problem.
I guess the problem is raised by a switch to another thread execution (e.g. as there is one https://github.com/me-no-dev/AsyncTCP/blob/master/src/AsyncTCP.cpp[line 221] starting a thread on any core). With a simple sketch (no AsyncTCP involved) this ir timing problem did not occur.

a) I would like to ask: is that the reason why this (or similar) pull request has not been accepted yet?
b) Has anybody already digged deeper into it? I think about asking Espressif...

makerMcl:
In addition to my previous comment: I have found a way solving this without workaround by modifying thread priority. 
I know there is a comment and link to a lib using ESP32-RMT, but it seems incomplete and does not implement all protocols.
Would you accept the solution with thread priorities? You can inspect at https://github.com/makerMcl/Arduino-IRremote. In case of acceptance I would apply the solution to all send()-methods.

bengtmartensson:
I have implemented ESP32 support (sending and writing) in [Infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino) version 1.1.0, based upon the content of this thread. (It contains dutycycle settable as a parameter by the user.) In a few hours, it will be available in the Arduino library manager, using the name "Infrared".

ArminJo:
Current version supports ESP32 sending

PR #541, opened 2017-11-30
Add XiaoMi1 support.

anders-liu:
Add supporting to Xiao Mi 1st generation IR remote controller, which is used on Xiao Mi Box I.

Only support to receiving data.

Xiao Mi uses a weird quaternary system to encode data, so it's a little bit different from other remote controllers.

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -76,8 +76,8 @@
 #define DECODE_PRONTO        0 // This function doe not logically make sense
 #define SEND_PRONTO          1
 
-#define DECODE_LEGO_PF       0 // NOT WRITTEN
-#define SEND_LEGO_PF         1
+#define DECODE_XIAOMI1         1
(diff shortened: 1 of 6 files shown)


IRremote.h
```diff
@@ -76,8 +76,8 @@
 #define DECODE_PRONTO        0 // This function doe not logically make sense
 #define SEND_PRONTO          1
 
-#define DECODE_LEGO_PF       0 // NOT WRITTEN
```
anders-liu: This is a mistake, must be something mismatch... Will revert soon.

IRremote.h
```diff
@@ -76,8 +76,8 @@
...
-#define SEND_LEGO_PF         1
```
anders-liu: Ditto.

ArminJo:
Please rebase the code if the protocol is still required.

PR #542, opened 2017-12-03
Sparkfun Pro Micro Support

use-sparingly:
Sparkfun Pro Micro boards are based on the Atmega32u4 chip. The pinout is similar to the Leonardo except that PC7 is not landed (no pin 13). However, Timer4 is available on O̅C̅4̅A̅/PC6/pin 5.

These changes rely on the board being configured in the Arduino IDE according to https://learn.sparkfun.com/tutorials/pro-micro--fio-v3-hookup-guide so that `-DARDUINO_AVR_PROMICRO` is set during compilation.

The changes include a small amount of cleanup around the OC1A pin definition where there was some unnecessary `if` block duplication.


z3t0:
Thank you so much for the contribution!

PR #543, opened 2017-12-04
Sparkfun Pro Micro and Leonardo Support (Beta branch)

alastairmccormack:
Sparkfun Pro Micro boards are based on the Atmega32u4 chip. The pinout is similar to the Leonardo except that PC7 is not landed (no pin 13). However, Timer4 is available on O̅C̅4̅A̅/PC6/pin 5.

These changes rely on the board being configured in the Arduino IDE according to https://learn.sparkfun.com/tutorials/pro-micro--fio-v3-hookup-guide so that -DARDUINO_AVR_PROMICRO is set during compilation.

Leonardo support has also been clarified as it had been sharing a few ifdefs with the Teensy 2.0 board leading to a confusion (by me at least) as to which pins it would use. This also allowed the 
travis-ci to build successfully. 

The changes include a small amount of cleanup around the OC1A pin definition where there was some unnecessary `if` block duplication.

This is the beta branch (and more concise) version of https://github.com/z3t0/Arduino-IRremote/pull/542


z3t0:
Just for clarification: Is the difference between #542  and this PR, the Leonardo Support

use-sparingly:
Hi @z3t0 , you're right, #543 is a better, more complete version for the beta branch. It was only after making #542 that I realised that the beta branch was the leading branch (IIRC, the Arduino IDE uses a release based off beta, right?), so prepared #543 and included new findings about Leonardo.

IMHO, #542 (master) and #543 (beta) can both be merged to their respective branches.

alastairmccormack:
#542 has been merged into beta. I'm going to close this PR and raise a new one for the additional Leonardo support, comments and simplified ifdefs

alastairmccormack:
Raised #694

PR #552, opened 2018-02-16
Update irPronto.cpp

Jvinicius06:
SENDPRONTO variable declared within IRremote.h
therefore IRremote.h should be called before it


ArminJo:
Included in the current version.

PR #557, opened 2018-03-16
Update IRremote.cpp

rainbow-cnay:
Looks like a symbolic error.  Right is :
  bool passed = ((measured_ticks >= TICKS_LOW (desired_us - MARK_EXCESS))
                && (measured_ticks <= TICKS_HIGH(desired_us + MARK_EXCESS)));


bengtmartensson:
One more comment: The original code does not really clearly communicate its purpose (which generated this PR). For example, the following had been much more self-explaining:

    int compensated_ticks = measured_ticks - MARK_EXCESS;
    bool passed = (compensated_ticks >= TICKS_LOW (desired_us))
                          && (compensated_ticks <= TICKS_HIGH(desired_us));

Issue #466, opened 2017-04-25
Having trouble with Pronto


beaglesparks:
**Board:** Nano
**Library Version:** 2.2.3
**Protocol:** Pronto

char DISH_GUIDE[] = "0000 0048 0000 000b 0016 021a 0016 009d 0017 00fa 0017 009b 0016 009b 0017 009c 0017 00fb 0017 00fb 0017 00fb 0017 00fb 0017 00fa";

...

       irsend.sendPronto(DISH_GUIDE, false, false);

I get the following error

undefined reference to `IRsend::sendPronto(char*, bool, bool)'

collect2.exe: error: ld returned 1 exit status

What am I doing wrong?

Thanks

z3t0:
Hmm might need more information.. no idea what collect2.exe is?

beaglesparks:
me either.  i've searched but can't find the reference.  Let's ignor it for now.  am i declaring DISH_GUIDE correctly? Also am i calling irsend.Pronto correctly?

beaglesparks:
I found collect2.exe.  it seems to be a secondary error related to the IDE https://forum.arduino.cc/index.php?topic=310950.0

bengtmartensson:
`collect2` is the main program of the loader. The original error messages means that the loader cannot find the said functions. 

The loader cannot find the function because no such function is not defined. In Pronto.cpp there is a function (line 73)

    bool sendPronto (char* s,  bool repeat,  bool fallback)

but defined that way, it is a global function, not a member function of IRsend (`IRsend::sendPronto`).

Quickly browsing through Pronto.cpp I see a number of problems...

z3t0:
Yep. I had deleted Pronto from the dev branch as well realising that either it is very poorly written or just written for a very specific use case a long time ago. That's about all I can offer at the moment though...

beaglesparks:
I have a Dish DVR i want to control and pronto codes for it.  do you have suggestions on what i need to do to try to control it? I can only find pronto codes for it, nothing else

bengtmartensson:
You can translate the Pronto form to raw microseconds using e.g. IrScrutinizer. The given one corresponds to 

+382 -9345 +382 -2727 +400 -4342 +400 -2692 +382 -2692 +400 -2710 +400 -4360 +400 -4360 +400 -4360 +400 -4360 +400 -4342

(Possibly you should dispose ot the signs.)

bengtmartensson:
[...]

PR #581, opened 2018-07-05
Fixing call in irPronto.cpp. Adding pronto example.

nseidle:
As it is the Arduino-IRremote library doesn't compile for Pronto. This fixes the function call and header arrangement. I also added an example but you can exclude it if it's too much.

### DIFF
--- a/examples/IRsendPronto/IRsendPronto.ino
+++ b/examples/IRsendPronto/IRsendPronto.ino
@@ -1,0 +1,55 @@
+/*
+   IRsendPronto: Deomonstrates how to send a Pronto code.
+   An IR LED must be connected to Arduino PWM pin 3.
+   Version 1.0 July, 2018
+   By Nathan Seidle @ SparkFun
+
+   If you have the remote for the device you are trying to control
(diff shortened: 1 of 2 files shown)


nseidle:
Looks like Pronto is going away - issue #466. I'm ok with that. As my example sketch shows, it sucks up a ton of progmem to send one code. According to bengtmartensson, looks like there may be ways to convert Pronto codes to raw micros. That method would probably work most of the time.

Closing this PR. Please re-open if you see any value in it.

bengtmartensson:
Just some comments:

> According to bengtmartensson, looks like there may be ways to convert Pronto codes to raw micros.

You can use IrScrutinizer to transform the Pronto form signals to the microsecond durations we all know. It can even, under some circumstances, generate a program skeleton for you.

Feel free to try [infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino) as an alternative to IRremote. It has (IMHO) many advantages over IRremote, including a working Pronto interface.

nseidle:
Very good. Thanks!

PR #593, opened 2018-10-14
iremote monitor

thecraftman:
Remote system monitor using arduino


ArminJo:
No relationship to IRremote and repo deleted. Was this PR meant for another project?

PR #598, opened 2018-12-24
API allowing disabling the timer interrupt.

ccoenen:
The ESP32 seems to be sensitive towards interrupts while an OTA update is running. After I call `enableIRIn()`, I can't get a successful OTA-Update any longer. The reason seems to be the timerinterrupt set up by IRRecv.

I now simply deactivate IRRecv once an update is being run like this:

```
ArduinoOTA.onStart([&]() {
  irrecv.disableIRIn(); // the new api contained in this pull request.
  Serial.println("Receiving Update");
});
```

### DIFF
--- a/IRremote.h
+++ b/IRremote.h
@@ -176,6 +176,9 @@ class IRrecv
 
 		void  blink13    (int blinkflag) ;
 		int   decode     (decode_results *results) ;
+#ifdef ESP32
+		void  disableIRIn( ) ;
+#endif
 		void  enableIRIn ( ) ;
 		bool  isIdle     ( ) ;
 		void  resume     ( ) ;
--- a/irRecv.cpp
+++ b/irRecv.cpp
@@ -116,6 +116,15 @@ IRrecv::IRrecv (int recvpin, int blinkpin)
 }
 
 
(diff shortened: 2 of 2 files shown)


IRremote.h
```diff
@@ -176,6 +176,9 @@ class IRrecv
 
 		void  blink13    (int blinkflag) ;
 		int   decode     (decode_results *results) ;
+#ifdef ESP32
+		void  disableIRIn( ) ;
+#endif
```
ccoenen: I limited both changes to the `ESP32`, because I can't test this on any other platform (also, so far nobody has needed this, and maybe it's not useful outside the esp32 world?)

ccoenen:
Is there any way to get this merged?

AnalysIR:
FYI: The repo owner (@z3t0) has not been active here in almost 1 year AFAIK, which may explain things.

ArminJo:
It is included in the current version 😀

bengtmartensson:
Presently `disableIRIn` is declared for all architechtures (`IRrremote.h`) but only defined for exp32 and the ones  using `USE_DEFAULT_ENABLE_IR_IN`.

Issue #359, opened 2016-09-06
Difference of pressing IRremote button for a long time or a shor time 


liang-jia:
**Board:** ARDUINO UNO
**Library Version:** 2.1.0
**Protocol:** Sony (if any)

HI,
 I have a problem  that the learning device sometimes can't succede learning when I press the IRremote button for a long time or a shor time ,can you tell me that it's diffrent for pressing long time or short ? Has the software some method to enhance the learning succeding rate for IRremote?

Thanks

z3t0:
I am not certain about all remotes but I know for my generic remote which is recognized as NEC, sends FFFF for repeat keys. 

IRremote doesn't necessarily handle this as different protocols may use different methods to signal long press (I'm not too sure), but you could easily program this as a part of your sketch.

I would recommend loading up IRRecvDumpV2 and take a look at what your remote does when holding down buttons

z3t0:
For example, the following is how the NEC works.
https://i.imgur.com/TcyNC24.png

The only way to determine how long it was pressed is to count up how many repeat signals are sent by the remote.

bengtmartensson:
IrScrutinizer can be used; it identifies a repeating portion of an ir sequence, and counts the number of occurances (on the plot: intro: red, repeat: blue, ending (when present): green).

Issue #28, opened 2013-05-02
NEC send repeat


alexandruy2k:
Using the NEC protocol the library can detect repeat codes and marks them as 0xFFFFFFFF, however if I need to send a repeat code I can't get it to work.

tried irsend.sendNEC(REPEAT, 0)
  irsend.sendNEC(REPEAT, 4) 
and  irsend.sendNEC(REPEAT, 32)

but none of these result in the NEC protocol repeat code
Is there a way to do it? If I get a quick hint at how to do it I can write it myself and submit it as a patch

z3t0:
Closed for inactivity. If you have been able to solve this issue please post how you did so, and it will be incorporated into the wiki.

BramVader:
I know this issue is somewhat old, but I found how to send repeat codes:

    unsigned int buf[3];
    buf[0] = 9000;  // Mark 9ms
    buf[1] = 2250;  // Space 2.25ms
    buf[2] = 560;   // Burst
    irsend.sendRaw(buf, 3, 38);

The reason why sendNEC() doesn't work is that normal messages have a 9ms mark followed by a 4.5ms space, followed by 32 mark/space bursts.

Galysh:
I also encountered such a problem
I think you can solve the problem as follows

Use code: irsend.sendNEC (REPEAT, 0) 


void  IRsend::sendNEC (unsigned long data,  int nbits)
{
	// Set IR carrier frequency
	enableIROut(38);
	// Header
	mark(NEC_HDR_MARK);
	**if (nbits == 0) space(NEC_RPT_SPACE);
		else space(NEC_HDR_SPACE);**
	// Data
	for (unsigned long  mask = 1UL << (nbits - 1);  mask;  mask >>= 1) {
		if (data & mask) {
			mark(NEC_BIT_MARK);
			space(NEC_ONE_SPACE);
		} else {
			mark(NEC_BIT_MARK);
			space(NEC_ZERO_SPACE);
		}
	}
	// Footer
	mark(NEC_BIT_MARK);
	space(0);  // Always end with the LED off
}

scipilot:
Both @BramVader and @Galysh are correct in the adjustment to the timing of the headers, but I couldn't get their code to work immediately. 

For the latter, I found sending nbits=0 into the function caused undefined behaviour due to the _negative_ left shift in the setup of the for loop - which should be skipped sending no data. So it was still sending one byte of data which invalidated the repeat command.

I detailed it here:
[...]

PR #609, opened 2019-02-20
Update ir_NEC.cpp

ufaman:
Correct REPEAT send

### DIFF
--- a/ir_NEC.cpp
+++ b/ir_NEC.cpp
@@ -26,16 +26,20 @@ void  IRsend::sendNEC (unsigned long data,  int nbits)
 
 	// Header
 	mark(NEC_HDR_MARK);
-	space(NEC_HDR_SPACE);
+	if (data==REPEAT)
+		space(NEC_RPT_SPACE);
+	else {
+		space(NEC_HDR_SPACE);
 
-	// Data
-	for (unsigned long  mask = 1UL << (nbits - 1);  mask;  mask >>= 1) {
-		if (data & mask) {
-			mark(NEC_BIT_MARK);
-			space(NEC_ONE_SPACE);
-		} else {
-			mark(NEC_BIT_MARK);
-			space(NEC_ZERO_SPACE);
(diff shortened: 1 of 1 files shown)


z3t0:
Would you mind writing a brief explanation of the change? Just for clarification, so that myself and other developers can make sure we understand the change :smile:

scipilot:
I was just about to submit the exact same patch, so I can explain what this change is. 

I mention the fix in https://github.com/z3t0/Arduino-IRremote/issues/28 with a link to a Stack Overflow about why the data loop should be suppressed, when sending the different repeat header.

a) the repeat signal has a different header (2500ms instead 4500ms) from normal keys

b) the repeat signal should send no data, but sending nbits=0 into this function causes unpredictable behaviour due to the resulting `<< -1` negative shift. So it can still send 1 bit, which breaks the signal.

It seems the FFFFFFFF code is virtual - it's not actually transmitted. I understood this from the image you linked in here: _Originally posted by @z3t0 in https://github.com/z3t0/Arduino-IRremote/issues/359#issuecomment-249346061_

The change in this PR fixes the repeat function for me on an Arduino Nano with a GVR TV (Australian Good Guys own-brand).

ArminJo:
Is just included manually.
Thanks 🥇

scipilot:
Sorry, but your manual edit only includes **one of the two** fixes. It has the conditional header, but doesn't prevent the intermittent 1 bit being sent. In this PR, the data sending for loop has also been put inside the ELSE for the non-repeat case.

`<< -1` is undefined behaviour in CPP, and I've indeed seen it behave intermittently, e.g. by writing tests run at different stack levels in the Arduino.

The for loop initialises `mask = 1UL << (nbits - 1);` In the case of a repeat `nbits` == 0 which leads to `<< -1`

So really the whole for data loop should be avoided if doing a repeat.

ArminJo:
Thanks for the QA!

Issue #462, opened 2017-04-24
Remove LED blinking


bengtmartensson:
Is the LED(-13) blink really a core functionality of the library? Or is it a debug feature that never was taken out? 

Disadvantages: 
* increases porting (and cleanup :-)) effort. 
* increased code size
* Locks up a LED that the user might like to use for some other purpose
* Can be annoying in deployment

Advantages:
* might be useful by initial debugging.
* Looks cool if you like it

Suggestion: take it out.

z3t0:
> Suggestion: take it out.

Agreed.

z3t0:
I have reopened this issue after considering that the LED might actually be useful to the end user. As such I would suggest rather than removing it we address the issues.

For example, it can be disabled on platforms that do not support it and we can add a relevant section in the documentation that explains how to disable the LED for advanced users.

bengtmartensson:
This issue has been resolved (by rejection). The issue thereby can and should therefore be closed.

PR #633, opened 2019-06-19
Set irparams.blinkflag to 1 when blinkpin is specified in IRrecv invocation

EasyG0ing1:
The only way to NOT have pin 13 be the blink pin is to specify your own pin for blinking when initially invoking IRrecv. However, the only way to set the flag for blinking is to call IRrecv.blink13 which sets the flag but it also sets pin13 as a blink pin.

I have an issue with a PCB that I etched that uses pin 13 for another function so setting the blinkflag via IRrecv.blink13 creates a conflict because it enables pin 13 for blinking.

Please, change the IRrecv invocation to set the blinkflag to 1 when a blinkpin is provided... why else would someone pass a different blink pin to the library unless they also wanted it to blink?

:-)


AnalysIR:
@EasyG0ing1 Yes indeed, after a **very quick** check on the Master, the code for 'blink' seems to be incomplete & will require someone to fix & submit it. (var types & logic wrong/incomplete)
As a workaround, it should be possible to handle your own blinking functionality as required in your own code. (I did not immediately see it using pin 13 by default, unless the blinkpin was specified. If it does, you should be able to overwrite it immediately in setup.)

@z3t0 fyi - I am not sure of the benefit of having this blink feature in the library at all, as it is easily implemented in user code, which would avoid issues like this.

EasyG0ing1:
> @EasyG0ing1 Yes indeed, after a **very quick** check on the Master, the code for 'blink' seems to be incomplete & will require someone to fix & submit it. (var types & logic wrong/incomplete)
> As a workaround, it should be possible to handle your own blinking functionality as required in your own code. (I did not immediately see it using pin 13 by default, unless the blinkpin was specified. If it does, you should be able to overwrite it immediately in setup.)
@AnalysIR - Ive been coding as a hobby since 1982. My primary profession is Network Engineering ... and I've never participated in any sort of Git collaboration effort submitting changes and what not ... HOWEVER, I did correct the library by changing the blinkflag to bool, and I corrected the problem that would occasionally cause the blink LED to stick to an on state. Also, when the user specifies a blink LED pin, it blinks by default without calling blink13(bool) and it can be toggled on or off with blink13(bool) [which I think should be renamed to blinkState(bool) so that the name is more generic and not specific to any given pin, but will still blink any hardwares default pin if it is set true without declaring a custom pin].

I have attached a zip file of the complete library as it exists now (by choosing download from https://github.com/z3t0/Arduino-IRremote) but with my changes and a PDF file called changes.pdf for your review (changed or added code is highlighted in red with before and after code clearly laid out).

> @z3t0 fyi - I am not sure of the benefit of having this blink feature in the library at all, as it is easily implemented in user code, which would avoid issues like this.

@AnalysIR - Well, you have blinking happening in ISR ... I am not able to determine how different the blinking of the LED would happen within Loop() vs ISR in the library, but my view on it is that I can write cleaner code when the library handles this for me, and it gives me confidence (perhaps incorrectly) that the blinking of the LED is actually happening in real time as decoding is happening vs me as a coder forcing it to happen in the Loop().  It may be a placebo, but it works and makes me feel good ...  lol
[Arduino-IRremote-master.zip](https://github.com/z3t0/Arduino-IRremote/files/3308912/Arduino-IRremote-master.zip)

bengtmartensson:
Led blinking was discussed in #462.

This project is simply stale. I have myself three pull requests, two years old... 61 pull requests altogether (including this one).

Alternatively, feel free to have a look at [Infrared4Arduino](https://github.com/bengtmartensson/Infrared4Arduino). It comes without blinking! PR welcome there!

AnalysIR:
@EasyG0ing1 I didn't see where it was implemented, so thanks for pointing out the ISR. It also looks like it was discussed previously with the conclusion it should be left in. I haven't checked your code implementation/fix, but it sounds reasonable. @z3t0 handles PRs for the library.

ArminJo:
Hi,
The PR is not longer required, since the blink handling has changed, and by defining CORE_LED0_PIN you can define your own blink LED pin.

PR #638, opened 2019-06-27
Fixed send NEC repeat command

sstefanov:
Fixed NEC send repeat command.
In existing code REPEAT space and marks are wrong.


sstefanov:
Duplicate with  #609

PR #648, opened 2019-08-13
Update IRremoteInt.h

Neo100998:


ArminJo:
Outdated, since parts of IRremoteInt.h were moved to boarddefs.h

PR #649, opened 2019-08-14
Add Check before defining used Timer

Popax21:
This allows the choosing of the used Timer by a Compiler flag (bsp. `-D IR_USE_TIMER1`)

### DIFF
--- a/boarddefs.h
+++ b/boarddefs.h
@@ -74,84 +74,116 @@
 
 // Sparkfun Pro Micro
 #if defined(ARDUINO_AVR_PROMICRO)
-	//#define IR_USE_TIMER1     // tx = pin 9
-	#define IR_USE_TIMER3       // tx = pin 5
-	//#define IR_USE_TIMER4_HS  // tx = pin 5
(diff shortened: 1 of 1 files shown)


bengtmartensson:
> choosing of the used Timer by a Compiler flag (bsp. -D IR_USE_TIMER1)

To my knowledge, there is no (at least not official) way of doing this. It is "officially" frown upon, see
 https://forum.arduino.cc/index.php?topic=487940.0  (just a random hit when I googled). (Note the arrogant tone "...you are not the target audience of the Arduino project and you should use a different IDE.  Arduino is mostly for people who don't even want to know what a compiler is..."

Popa21:
@bengtmartensson I know that, but this library is (at least to my knowledge) not just targeted at the Arduino IDE (I'm using it with PlatformIO). And the patch doesn't change the behaviour of the library if no timer is specified, so there would be no effect for the normal user. But if more advanced users (which are probably using a more advanced IDE) want to change the setting, they can do that without changing the source code of the library.

PR #658, opened 2019-10-23
Add support for Mitsubishi protocol sending

JakubKohoutek:
This PR enables sending the IR code using Mitsubishi protocol as described here: https://sbprojects.net/knowledge/ir/xsat.php

I verified that the function works on a real device.


ArminJo:
Sorry, but introducing a new timing is not possible for backward compatibility

PR #675, opened 2020-02-09
[BREAKING] Fix NEC: LSB first and support for REPEAT

lionello:
* Many online docs quote NEC codes but they don't work because they should be transmitted LSB first. Fixed both `sendNEC` and `decodeNEC` to be LSB first. *This can break existing code!*
* Added support for `REPEAT` in `sendNEC`


lionello:
I agree with not merging this, but suggest to leave it so other's can find it when their codes don't match what the manual shows.

ArminJo:
@lionello 
since the beta branch is going to be deleted, please open an issue for this valuable information.
Thanks.
I will therefore now close the PR.

PR #681, opened 2020-03-19
Resolves undefined reference error when calling sendPronto

raudette:
There may be a better way to fix this issue, but here's what I've done.

When you follow the example code like IRSendDemo.ino, you call a function like this:
irsend.sendSony

This wasn't working for sendPronto. When I did an irsend.sendPronto, the Arduino IDE would generate the following error:
undefined reference to `IRsend::sendPronto(char*, bool, bool)'

There were two issues.  IRremote.h defines the send function to return void, and the sendPronto in irPronto.cpp returns bool.

Second, just the way the TEST code in sendPronto was setup, IRremote.h wasn't getting included, and it was missing the:
#define SEND_PRONTO          1

### DIFF
--- a/irPronto.cpp
+++ b/irPronto.cpp
@@ -1,24 +1,6 @@
-#define TEST 0
-
-#if TEST
-#	define SEND_PRONTO        1
-#	define PRONTO_ONCE        false
-#	define PRONTO_REPEAT      true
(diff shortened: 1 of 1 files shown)


ArminJo:
Pronto is intentionally disabled. And we have actually 2 versions to choose 😩 .

PR #689, opened 2020-05-20
Fixes core panic when using WiFI and IRremote on ESP32

dspverden:
ESP32 (with Arduino) runs into core panic when IRremote is used together with WiFi. This is caused by the switch block in IRTimer(). Chaging it to if statements fixes the issue.

### DIFF
--- a/IRremote.cpp
+++ b/IRremote.cpp
@@ -124,7 +124,7 @@ int  MATCH_SPACE (int measured_ticks,  int desired_us)
 //   Gap width is recorded; Ready is cleared; New logging starts
 //
 #ifdef IR_TIMER_USE_ESP32
-void IRTimer()
+void IRAM_ATTR IRTimer()
 #else
 ISR (TIMER_INTR_NAME)
 #endif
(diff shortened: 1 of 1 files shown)


z3t0:
I think this is a critical fix and I would like to merge it in. However, I also don't have a way to test it. If some one else with the same hardware configuration can replicate the issue and then confirm this fix works for them then I can move forward with the merge.
@dspverden This is great thanks!

I want to merge this in but also have some concerns around the development of this library and workflow around merging PRs into master. The main thing is I want to get bug fixes into master, but also want to make sure the fixes can be replicated by other users since I can't confirm it myself. This may slow down merging however so if anyone has other ideas for the workflow then I'd love to hear them!

fgietzen:
I can confirm there is a problem with using the Arduino's WiFi class. The ESP32 simply crashes with `Cache disabled but cached memory region accessed`

With this fix everything works as intended.

*Note:* I applied this patch to the latest version on master. Every state must renamed to for example `IR_REC_STATE_IDLE.`

I am quite confused why the existing `switch-case` code is not working and must be replaced by `if` blocks. In the end, it should do the extact same thing but it doesn't. I ended using only the if-variant.

ArminJo:
It is due to the ESP32 compiler bug espressif/esp-idf#1552.
Merged manually

PR #690, opened 2020-05-24
Add bosewaveradio support

uvotguy:
Add support for a Bose Wave Radio CD player.

I had no luck in finding a spec for the Bose protocol; however, I have a working remote and player.
 I believe the IR LED in my remote is getting weak - iIt only works at close range even with a fresh battery.  I wanted to make my own remote control before it died, so here I am!

I captured signals from each button with a simple TSOP38238 IR Receiver Module (a.k.a. phototransistor) and an oscilloscope.  Once I got the protocol down, I captured a number of signals and averaged the timing for the relevant waveform components.

My new Arduino based remote works well even at a distance.

FYI I'm new at this Open Source thing, so please let me know if I've stepped on any toes, or if something can be improved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 *.un~
 *.sublime-project
 *.sublime-workspace
-ir_BoseWave.cpp.bak
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.un~
 *.sublime-project
 *.sublime-workspace
+ir_BoseWave.cpp.bak
--- a/IRremote.h
+++ b/IRremote.h
@@ -79,6 +79,9 @@
 #define DECODE_LEGO_PF       0 // NOT WRITTEN
 #define SEND_LEGO_PF         1
 
+#define DECODE_BOSEWAVE      1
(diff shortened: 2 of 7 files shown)


ArminJo:
Just one question.
Why did you not find/use the BOSE protocol [here](https://github.com/ukw100/IRMP#list-of-protocols)?

uvotguy:
Thanks for the reply, @ArminJo.  I wasn't aware of IRMP.  LIRC came up in my search, but the timing didn't match my measurements.  Your m'ments match mine nicely.  Since IRMP dev is more active, I'll give it a try.  Thanks!

Quick question for you. Was your timing done empirically, or did you work from a spec?

ArminJo:
I guess the timing was done empirical, I am not the developer of the software.
If you have problems with IRMP do not hesitate to open an issue there 😀 .

BTW your code seems quite clean, so maybe it will be merged later.

uvotguy:
Thank you.  I always try to emulate the project style.  I hope somebody can find use for it if they're wed to IRremote.

Can't wait to try IRMP!!!!

PR #702, opened 2020-07-10
Pronto implementation

bengtmartensson:
This PR implements `IRrecv::dumpPronto` and `IRsend::sendPronto`. There is a new example that demonstrates `sendPronto` (both for signals in RAM, PROGMEM, and the F()-form). The example `IRrecvDumpV2` has been extended to generate output from `dumpPronto`. 

[Reference](http://www.harctoolbox.org/Glossary.html#ProntoSemantics)


examples/IRrecvDumpV2/IRrecvDumpV2.ino
```diff
@@ -9,7 +9,7 @@
 #if defined(ESP32)
 int IR_RECEIVE_PIN = 15;
 #else
-int IR_RECEIVE_PIN = 11;
+int IR_RECEIVE_PIN = 5;
```
ArminJo, line 12: Why this change?
It is inconsistent with all other examples then.

bengtmartensson: Just forgot to set it back. No reason.

ArminJo: Thanks

ArminJo:
Hi Bengt,
Thanks for the code, I cant realy test it, and do not know much about Pronto, but it looks fine so far.
Would you like to sleep one night before merging (my approach) or should I merge the PR now?
Best Regards
Armin

bengtmartensson:
Armin, take your time. Best would be if someone can write a review,  @probonopd ??

ArminJo:
I need no time 😀  I make what you propose!
My experience is, that the developer (mostly me) should wait for 24 hours before releasing a bigger change.
@probonopd It would be fantastic if you can review the code.

probonopd:
When it comes to reviewing C++ I'm the wrong guy, sorry, but I really trust what @bengtmartensson is doing.

PR #709, opened 2020-08-11
Arduino Zero support?

gavron04:
Hello. Now Im using 32u4 atmel (arduino micro), but it has only 32kb flash. I found nice new Arduino - Zero. Its possible to make IRremote work with Zero?

Have a good day

Regards


ArminJo:
There is SAMD support in the current version.
You can also try [IRMP](https://github.com/ukw100/IRMP), it also has SAMD support.

PR #716, opened 2020-10-10
Fix multiplication overflows

JohnChen0:
Multiplying two 16-bit integers can yield a result that does not fit
in a 16-bit integer, causing overflow and result in wrong result.
Fixing a couple of places that this can happen.

### DIFF
--- a/src/ir_Sanyo.cpp
+++ b/src/ir_Sanyo.cpp
@@ -38,7 +38,7 @@ bool IRrecv::decodeSanyo() {
 #endif
 
 // Initial space
-    if ((results.rawbuf[offset] * MICROS_PER_TICK) < SANYO_DOUBLE_SPACE_USECS) {
+    if (results.rawbuf[offset] < SANYO_DOUBLE_SPACE_USECS / MICROS_PER_TICK) {
         //Serial.print("IR Gap found: ");
         results.bits = 0;
         results.value = REPEAT;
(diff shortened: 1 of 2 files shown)


AnalysIR:
Note:
an overflow should only happen, if the data is invalid. So in both cases the result is an error/incorrect.
(for valid timings/settings there should not be an overflow)

AnalysIR:
forot to mention earlier...

`if (results.rawbuf[offset] < (SANYO_DOUBLE_SPACE_USECS + MICROS_PER_TICK/2) / MICROS_PER_TICK) ) {`

vs
`if (results.rawbuf[offset] < SANYO_DOUBLE_SPACE_USECS / MICROS_PER_TICK) {`

should provide better rounding for integer division, without any additional runtime overhead.

ArminJo:
But the values are 500 and 800, so no rounding error. And the values look like someone guessed them....

AnalysIR:
OK
:)

JohnChen0:
> Note:
> an overflow should only happen, if the data is invalid. So in both cases the result is an error/incorrect.
> (for valid timings/settings there should not be an overflow)

To clarify, the overflow can cause valid commands for other protocols to be incorrectly recognized as Sony or Sanyo repeat codes. In one particular case I encountered, my IR receive has received a JVC code that happens to have results.rawbuf[0] == 6554. Translating that into microseconds should yield 6554*50=327700, but after 16-bit overflow the multiplication yields 20, causing the if condition to be true, and the JVC code incorrectly recognized as Sanyo/Sony repeat code (since decodeSony and deocdeSanyo are called before decodeJVC).

PR #717, opened 2020-10-10
Fix bit counting in decode JVC

JohnChen0:


ArminJo:
Thanks for pointint this out, but you forgot some protocols with the same error.

PR #727, opened 2020-11-22
* dumpPronto now takes a String object as argument

iosifpeterfi:
Hello,

This PR extends dumpPronto. It allows the function to take a char pointer as first argument, then fills the buffer with the Pronto hex information.

malloc() and free() and handled outside the call. If there's a way to handle this better we can update it.

Regards,
Iosif


ArminJo:
Thanks again 🥇 A good improvement 👍 
But it is hard to understand and not very common using `*&varCha`r
Can you convert it to `size_t dumpSequence(char* aStringBuffer,  size_t aBufferLength, const volatile uint16_t *data, size_t length, uint16_t timebase)` to make it more readable and to avoid buffer overflows.
This would be great!!!
Armin

iosifpeterfi:
Added buffer size checks.
Updated the function type as bool, to return true when done and false when buffer size is too small.

If I convert it to a regular pointer, the value is lost and the data is copied in memory. I want to reuse this variable defined externally, the same way I use the Serial console.  This implementation dumps directly to the pointer referenced in the parameter.

ArminJo:
> same way I use the Serial console

I do not understand, because the signature of Serial and Print is:
**size_t** println(const String &s); 
You can add the size_t return value to your pointer and you have the end of the string, or check the value against the requested length.

iosifpeterfi:
I think this is much easier to use and implement.

It uses as String() object instead of a *&char pointer.

ArminJo:
Looks very promising 👍. But do you know the String.length() method?
And please can you insert a test call at the end of the loop in IRreceiveDumpV2.cpp? Or do you have an example for using the new function 😀 ?

iosifpeterfi:
Hello Armin,

The function allows to add more data to an existing string.
I know the String.legth() but this returns the total length of the whole string, not the length of the data that was appended to it.

The reason why I wrote is to upload this data in the Pronto Hex format. However implementing this in a demo would have too many requirements. What we can do in a demo is to feed the codes back into prontoSend, similar to IRrecord.

Would that make sense?

ArminJo:
> The function allows to add more data to an existing string.
> I know the String.legth() but this returns the total length of the whole string, not the length of the data that was appended to it.

Valid argument! 😀 

As demo I require just a line or two which shows how to convert the raw data to a "String" and then print it. (which can be used for internal testing)
It might be trivial...

iosifpeterfi:
Updated. Sorry for the commit spam, I had no luxury to compile this locally.

bengtmartensson:
Just a comment. Code like

     String ProntoHEX = ...;
     IrReceiver.dumpPronto(&ProntoHEX);
     Serial.println(ProntoHEX);

is not good coding for an embedded system with limited memory.  The reason is that a pretty large object is constructed. For a typical signal, say NEC1 with one repeat, the string is 380 characters long, the String object is certainly at least as long (in bytes), and that is _a lot_ for a processor as ATMega328P (as in the Arduino Uno and Mega). Even though I have not tried it, I severely doubt that the modified IRreceiveDumpV2 will run on an Uno or Nano.

Using the `IrReceiver.dumpPronto(Serial)` (or is it IrReceiver.dumpPronto(&Serial)?)  does not have this problem, since it writes every character as it is computed. There is/was a reason why my original did not implement a String/char* generating function.

iosifpeterfi:
@bengtmartensson I agree, this needs to run on a embedded system with more memory, used IoT (ESP32, Particle, etc). I think is best to create a complete separate demo file for this function, this way the dumpProto stays the same and does not break the current purpose of dumpProto.

ArminJo:
IMHO its OK to have the dump to String functions available.
Nobody is forced to use them and then no additionally program space is used.
But you/we should add a remark, how many RAM/program space is used on an AVR by converting one code. If you want I can do this for you.

iosifpeterfi:
The memory space used pronto results.rawlen. It is results.rawlen + 25 bytes. It can be 400 bytes long or more. Some appliances like air conditioners use more than 1Kb. This amount of SRAM when the function runs. The regular rules apply and the the String object uses the SRAM until is deleted.

Normally the loop deletes the string object automatically, however I added the delete call in the code just to make it more obvious when someone copies it in a different scenario.

iosifpeterfi:
Not sure why the delete does not work as is supposed to so I reverted the changes from the last couple of commits.

iosifpeterfi:
Is really weird how the build fails with the remove() and delete statements. I rechecked a local compile and is fine so I'm not sure what is missing.

ArminJo:
@iosifpeterfi Thank you very much 🥇 .
I documented it and removed the delete which is not required for a local variable.

PR #742, opened 2020-12-27
Fix Sony 20 bit decoding (return the number of bits decoded)

leres:
Testing done with a Sony Cybershot RMT-DSC1 remote

### DIFF
--- a/src/ir_Sony.cpp
+++ b/src/ir_Sony.cpp
@@ -51,6 +51,7 @@ void IRsend::sendSony(unsigned long data, int nbits) {
 #if DECODE_SONY
 bool IRrecv::decodeSony() {
     long data = 0;
+    uint16_t bits = 0;
     unsigned int offset = 0;  // Dont skip first space, check its size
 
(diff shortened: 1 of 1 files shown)


leres:
I just found a Sony RM-X201 remote that sends 15 bit codes.

ArminJo:
You expect more than 255 bits for Sony???

leres:
> You expect more than 255 bits for Sony???

I assume you're referring to:

    `uint16_t bits = 0;^M`

I made it the same size as the bits member in the decode_results struct.

While the nbits argument to IRsend::sendNEC() and IRsend::sendRC6 are uint8_t, all others are int (16 bits). Local bits/nbits variables in irPronto.cpp, ir_RC5_RC6.cpp, and ir_Sharp_alt.cpp are all 16 bit (either unsigned int or uint16_t).

ArminJo:
Thanks!
I haven't payed attention to this, I am going to change it glabally after the merge.

PR #816, opened 2021-03-03
Added IR Receiver support for STM32 boards based on STM32Duino 1.9.0

pmalasp:
Based on the HardwareTimer library in STM32Duino 1.9.0.

Tested only on STM32F103 boards (bluepill) with TSOP4838 IR Receiver.

STM32 timer: TIM4, can easily be changed to any other existing HW timer.


ArminJo:
Thanks for your valuable contribution 👍 
Since we are in a major refactoring, and some parts like the Continuous Integration are missing in your PR, I would propose to include your changes manually and add you to the contributors list.
For the list, you can specify another name than pmalasp if you like.
Is this OK for you?

pmalasp:
That is absolutely fine with me :-) !

Paolo Malaspina is the real name .

Best regards

Paolo

PR #902, opened 2021-10-14
Update README.md

gaoyaoxin:
Added some information about sending.


gaoyaoxin:
> Thanks, but there are two diffs I cannot see the reason for. Is it OK If I merge your changes manually, and not by pull request?

Sure.
"We’ve detected the file encoding as ISO-8859-1. When you commit changes we will transcode it to UTF-8." popped up while modifying, so the last two might be changed somehow.

ArminJo:
Thanks, I dedicated the last commit to you 😀

PR #903, opened 2021-10-16
Fix include in multi files

Smeagolworms4:
Fix a "multiple definition" error and raise if you include the library in more than one file.

example:

MyLib.h:

```
#ifndef MY_LIB_H
#define MY_LIB_H

#include "IRremote.h"

class MuClass {

};

#endif
```

MyLib.cpp:

```
#include "MyLib.h"

```

main.cpp:

```

#include "MyLib.h"

```


ArminJo:
Plese explain the benfits of this change a bit more.

Smeagolworms4:
The structure of the library does not allow it to be included in several files or in .h and .cpp split files.


Without my modification if we try to include the library from multiple files or from a library with separate files. Then we got a compilation error.

Mutiple definition on all objects in hpp files (or cpp.h before)

Hpp files cannot be used to be exposed in a .h file

They can only be used for private parts of the library

My modification is just to split the exposed hpp files into .h and .cpp files

ArminJo:
And why don't you just include the IRremoteInt.h file in all files which does not contain main()?

Smeagolworms4:
This example d'ont build with actual code of library.

 main.cpp

```cpp
#include <MyLib.h>

void setup() {
    MyClass myClass();
}
void loop() {}

```
MyLib.h

```cpp
#ifndef MY_LIB_H
#define MY_LIB_H

#include <IRremote>

class MyClass {
    public:
        MyClass();
}

#endif
```

MyLib.cpp

```cpp
#include "MyLib.h"

MyClass::MyClass() {}

```

ArminJo:
And why don't you just include the IRremoteInt.h file in all files which does not contain main()?

ArminJo:
Use #include <IRremote.h> only in the .ino file (the file containing main()) and use #include <IRremoteInt.h> in all other files.

Smeagolworms4:
This therefore requires you to always load the lib from the ino file (or main.cpp).

In my case there is no longer any distinction to be made and other libraries can load it.

Only one entry point remains on the library.

But it is true that using IRremoteInt.h in the sub lib will work I think

ArminJo:
Any thoughts how you handle setting of macros like USE_NO_SEND_PWM or EXCLUDE_EXOTIC_PROTOCOLS in your code?

Smeagolworms4:
Place it in the main.cpp or in the MyLib.h library as needed with an ifndef if we want to overload in the lib all while keeping it modifiable

Smeagolworms4:
Version 1

 main.cpp

```cpp

#define USE_NO_SEND_PWM 1

#include <MyLib.h>

void setup() {
    MyClass myClass();
}
void loop() {}

```
MyLib.h

```cpp
#ifndef MY_LIB_H
#define MY_LIB_H

#include <IRremote>

class MyClass {
    public:
        MyClass();
}

#endif
```

MyLib.cpp

```cpp
#include "MyLib.h"

MyClass::MyClass() {}

```



Version 2

 main.cpp

```cpp

#include <MyLib.h>

void setup() {
    MyClass myClass();
}
void loop() {}

```
MyLib.h

```cpp
#ifndef MY_LIB_H
#define MY_LIB_H


#ifndef USE_NO_SEND_PWM
#define USE_NO_SEND_PWM 1
#endif 

#include <IRremote>

class MyClass {
    public:
        MyClass();
}

#endif
```

MyLib.cpp

```cpp
#include "MyLib.h"

MyClass::MyClass() {}

```

ArminJo:
Please make an PR with proper documentation, which covers all use cases and run for all examples

PR #914, opened 2021-11-18
Declare header files for PlatformIO

ivankravets:
Hi,

Please accept this PR that adds header files according to https://docs.platformio.org/en/latest/librarymanager/config.html#headers

Developers will see them soon in the new version of PlatformIO Package Registry.

Thanks in advance!

### DIFF
--- a/library.json
+++ b/library.json
@@ -1,6 +1,6 @@
 {
   "name": "IRremote",
-  "keywords": "infrared, ir, remote",
+  "keywords": "communication, infrared, ir, remote",
   "description": "Send and receive infrared signals with multiple protocols",
   "repository":
   {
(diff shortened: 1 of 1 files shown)


ArminJo:
No problem, but one question.
As you can see, the IRremote.h is a redirect to IRRemote.hpp, in order to be backwards compatible.
I changed the whole library to be included as sources (*.hpp)  in order to be able to define macros valid for all sources. This is the only way for the Arduino "IDE" . https://github.com/Arduino-IRremote/Arduino-IRremote#compile-options--macros-for-this-library and https://github.com/Arduino-IRremote/Arduino-IRremote#supported-ir-protocols 

So why not specify IRremote.hpp as header or if you require a real header, IRremoteInt.h? See https://github.com/Arduino-IRremote/Arduino-IRremote#requiring-irremoteh-in-multiple-files--avoiding-multiple-definition-error
What is your advice?

Thanks
Armin

ivankravets:
Thanks, updated!

PR #928, opened 2021-12-19
Update IRTimer.hpp for Teensy boards

PaulStoffregen:
Adds timer defines for Teensy 4.0 & 4.1, and fixes timer defines for other Teensy boards.


ArminJo:
Thanks a lot 🥇

PR #955, opened 2022-01-23
Add pi pico support

tfry-git:
Apparently, earlier versions of the arduino-pico core were mbed-based. This appears to be no longer the case. At least not for the popular core, here: https://github.com/earlephilhower/arduino-pico . This patch adds support, again.

Only receive was actually tested.

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,6 +382,7 @@ ATtiny and Digispark boards are only tested with the recommended [ATTinyCore](ht
 - Sparkfun Pro Micro
 - Nano Every, Uno WiFi Rev2, nRF5 BBC MicroBit, Nano33_BLE
 - BluePill with STM32
+- RP2040 based boards (Raspberry Pi Pico, etc.)
 
 We are open to suggestions for adding support to new boards, however we highly recommend you contact your supplier first and ask them to provide support from their side.<br/>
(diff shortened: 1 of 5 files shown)


tfry-git:
Note: I have no idea what's supposed to have changed on line 1096 of IRTimer.hpp.

tfry-git:
Ok, now cleaned up. Part of the problem seems to have been that I had used the online editor for my initial tinkering. But also you seem to have made a force push while I was preparing my patch.

ArminJo:
Thanks again for your contribution.
Who does the CI and documentation work?

tfry-git:
> Who does the CI and documentation work?

Added documentation. I hope, I did not overlook anything. I have no idea, how to add CI, sorry. 

> Who does the sending tests?

Whoever first has a need for it? Seriously, I do not have the hardware setup for this, but no dedicated code was added, this is an arduino core, the board is fast, and nearly unrestricted WRT to pinout, so I do not really expect issues (famous last words, but still).

> Who adapts the examples?

You mean PinDefintionsAndMore.h? I can do that, although I'm wondering, if the current approach isn't a bit over-explicit:
Many of the more recent boards are mostly unrestricted wrt pin usage, so it's no surprise, many different boards specify the same pinout. library.properties already provides for a warning for unsupported boards. Hence I'd suggest to remove the warning in PinDefinitionsAndMore.h, remove any #ifdefs that simply re-state these defaults, and all the defaults, silently.

Your call, of course.

PR #963, opened 2022-01-30
Checks DECODE_DISTANCE before including protocols

same31:
Similar to the already checked `defined(DECODE_HASH)`, I've added a check on `defined(DECODE_DISTANCE)` if only pulse width / distance is required.


ArminJo:
Thanks 👍

PR #967, opened 2022-02-03
Declare the variable 'extern'

skb666:
It is recommended to change the variable to user - defined to increase code robustness


ArminJo:
You see the Problem?

PR #989, opened 2022-05-15
ESP32 Fix send and receive

htmltiger:
bug: receive only worked if send was not used, once any code sent it stopped receiving.
disabled interrupt  was not re-enabled in `IrReceiver.resume()` and calling `IrReceiver.enableIRIn()` after `IrSender.sendNEC()` was crashing


aakoch:
Does this fix the "HW TIMER NEVER INIT ERROR" I'm getting? If so, 👍🏻

PR #1026, opened 2022-09-11
Fix missing Magiquest declaration in Protocol list

recordandmix:
Supposedly “If no protocol is defined, all protocols are active.”  Code at line 97-102 of IRremote.hpp first establishes that NO_DECODER is not active, then, one by one, establishes that none of the other individual protocols have been declared active, either.  At that point, all protocols are supposedly activated.  However, the reference to defined(DECODE_MAGIQUEST) is missing.  As such, if you try to declare #define DECODE_MAGIQUEST in your code,  ALL protocols will be activated, not just Magiquest.  The missing reference needs to be added to line 102, so that code which seeks to only decode Magiquest commands will work correctly.

### DIFF
--- a/src/IRProtocol.h
+++ b/src/IRProtocol.h
@@ -8,7 +8,7 @@
  ************************************************************************************
  * MIT License
  *
- * Copyright (c) 2020-2022 Armin Joachimsmeyer
+ * Copyright (c) 2020-2021 Armin Joachimsmeyer
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


ArminJo:
Thank you very much for reporting. 🥇 
I fixed it manually, since your PR covers other lines of the code.

recordandmix:
Thanks!  Long time user, first time PR-creator.  When I tried to make the fix in the master branch, it said I was already 3 commits behind.  I couldn't figure out how to roll that back, at which point I just decided to kinda tweak the description to make it a bug report.

PR #1028, opened 2022-09-28
Prevent long delay caused by overflow when frame duration < repeat period

sjahu:
I noticed that when the repeat period for `sendPulseDistanceWidthFromArray` or `sendPulseDistanceWidth` happens to be shorter than the time it actually took to send a frame, the calculation for the delay overflows. This leads to an indefinite/very long delay, basically hanging the program "forever".

This PR adds a guard to prevent this overflow. Rather than hanging, when the repeat period is shorter than the actual frame duration, the program will proceed immediately to the next repetition. This doesn't feel totally correct since it means the period will be wrong, but IMO it's a better and more forgiving result than the status quo.

NB: Calling these methods with such a short repetition period is probably a mistake (as it was in my case). Still, this fix might save some future users some trouble.

### DIFF
--- a/src/IRSend.hpp
+++ b/src/IRSend.hpp
@@ -399,7 +399,10 @@ void IRsend::sendPulseDistanceWidthFromArray(uint_fast8_t aFrequencyKHz, unsigne
         tNumberOfCommands--;
         // skip last delay!
         if (tNumberOfCommands > 0) {
-            delay(aRepeatPeriodMillis - (millis() - tStartOfFrameMillis));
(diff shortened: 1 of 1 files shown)


ArminJo:
Hi Stephen,
thank you very much for your PR. 🥇 
I merged it manually, since I slightly changed the code and applied the patch to all 4 places, even to the two ones, where it should not be neccesary. But this makes the code more consistent and for my surprise the code size does not increase with your patch 😀 .
Best regards
Armin

sjahu:
🎉

PR #1077, opened 2023-01-13
Fix attiny x4 pinout for fast pin read and write

MathiasMoog:
Fix Pin numbering for ATtiny 84. The typical clockwise pinout, i.e. ATTinyCore https://github.com/SpenceKonde/ATTinyCore assigns Arduino D8 to PB2.  
Assigning D8 to PB0 is more common for the 167 types. 
Add another decision on the pin mapping to distinguish both types.
Tested on ATtiny 84 with ATTinyCore and TinyIRReceiver.


ArminJo:
Hi Mathias,
thanks for reporting 🥇 .
I added a bit and commited it. I hope this is OK for you.
Armin

PR #1158, opened 2023-08-14
Update README.md

lemio:
Added code to reverse the from MSB to LSB and vice versa


ArminJo:
But thanks anyway, I extended the readme a bit 😀 .

PR #1199, opened 2024-02-12
Fix stringop-truncation warning with strncpy

alvarop:
While building my project, I noticed the following warning from the ir_Pronto.hpp **sendPronto** implementation.

```
In file included from /Users/alvarop/.platformio/packages/framework-arduinoteensy/cores/teensy4/WProgram.h:42,
                 from /Users/alvarop/.platformio/packages/framework-arduinoteensy/cores/teensy4/Arduino.h:6,
                 from .pio/libdeps/teensy41/IRremote/src/IRremoteInt.h:36,
                 from .pio/libdeps/teensy41/IRremote/src/IRRemote.hpp:256,
                 from src/main.cpp:4:
.pio/libdeps/teensy41/IRremote/src/ir_Pronto.hpp: In member function 'void IRsend::sendPronto(const __FlashStringHelper*, int_fast8_t)':
/Users/alvarop/.platformio/packages/framework-arduinoteensy/cores/teensy4/avr/pgmspace.h:70:35: warning: 'char* strncpy(char*, const char*, size_t)' output truncated before terminating nul copying as many bytes from a string as its length [-Wstringop-truncation]
   70 | #define strncpy_P(a, b, n) strncpy((a), (b), (n))
      |                            ~~~~~~~^~~~~~~~~~~~~~~
.pio/libdeps/teensy41/IRremote/src/ir_Pronto.hpp:193:5: note: in expansion of macro 'strncpy_P'
  193 |     strncpy_P(work, reinterpret_cast<const char*>(str), len);
      |     ^~~~~~~~~
/Users/alvarop/.platformio/packages/framework-arduinoteensy/cores/teensy4/avr/pgmspace.h:65:27: note: length computed here
   65 | #define strlen_P(s) strlen((const char *)(s))
      |                     ~~~~~~^~~~~~~~~~~~~~~~~~~
.pio/libdeps/teensy41/IRremote/src/ir_Pronto.hpp:191:18: note: in expansion of macro 'strlen_P'
  191 |     size_t len = strlen_P(reinterpret_cast<const char*>(str));
      |                  ^~~~~~~~
```

Since **strncpy** doesn't add a trailing zero if the string is the maximum length, the `work` string might not be NULL terminated. Using the `sizeof(work)` gets rid of this issue since work will always be one larger than the length of the string.

### DIFF
--- a/src/ir_Pronto.hpp
+++ b/src/ir_Pronto.hpp
@@ -174,7 +174,7 @@ void IRsend::sendPronto(const char *str, int_fast8_t aNumberOfRepeats) {
 void IRsend::sendPronto_PF(uint_farptr_t str, int_fast8_t aNumberOfRepeats) {
     size_t len = strlen_PF(str);
     char work[len + 1];
-    strncpy_PF(work, str, len);
+    strncpy_PF(work, str, sizeof(work));
     sendPronto(work, aNumberOfRepeats);
 }
 
@@ -182,7 +182,7 @@ void IRsend::sendPronto_PF(uint_farptr_t str, int_fast8_t aNumberOfRepeats) {
 void IRsend::sendPronto_P(const char *str, int_fast8_t aNumberOfRepeats) {
     size_t len = strlen_P(str);
     char work[len + 1];
-    strncpy_P(work, str, len);
+    strncpy_P(work, str, sizeof(work));
     sendPronto(work, aNumberOfRepeats);
 }
 #endif
@@ -190,7 +190,7 @@ void IRsend::sendPronto_P(const char *str, int_fast8_t aNumberOfRepeats) {
 void IRsend::sendPronto(const __FlashStringHelper *str, int_fast8_t aNumberOfRepeats) {
     size_t len = strlen_P(reinterpret_cast<const char*>(str));
     char work[len + 1];
-    strncpy_P(work, reinterpret_cast<const char*>(str), len);
+    strncpy_P(work, reinterpret_cast<const char*>(str), sizeof(work));
     return sendPronto(work, aNumberOfRepeats);
 }
 


ArminJo:
??? what is the problem?

alvarop:
I may be mistaken, but this is how I see it:
`char work[len + 1];` is stack allocated and uninitialized, so we don't know what the value of `work[len]` is. 
`strncpy_P(work, reinterpret_cast<const char*>(str), len)` will copy the string from `str` to `work`, but it will not copy the NULL character at the end, because `len = strlen_P(reinterpret_cast<const char*>(str));`, so it will stop copying when it reaches `len`.

From [man strncpy](https://linux.die.net/man/3/strncpy):
> The strncpy() function is similar, except that at most n bytes of src are copied. Warning: If there is no null byte among the first n bytes of src, the string placed in dest will not be null-terminated.
>If the length of src is less than n, strncpy() writes additional null bytes to dest to ensure that a total of n bytes are written.

The function then calls `IRsend::sendPronto(const char *str, int_fast8_t aNumberOfRepeats)` which itself does `size_t len = strlen(str) / (digitsInProntoNumber + 1) + 1;`

If `work[len]` is not 0, `strlen(str)` in `sendPronto` will return a value larger than the actual length of the string. Since you're dividing by `digitsInProntoNumber + 1`, as long as there's a 0 within 5 bytes of `work[len]`, everything will work as expected. If the memory in those addresses doesn't have 0's, then `strlen(str)` will be larger than expected and additional "garbage" data may be transmitted.

I'm not actually using sendPronto, but I like having no warnings on compilation to prevent missing potential problems in the future. 😅

ArminJo:
Thank you very much. 🥇 
Now I see the problem you address.
But I would prefer the solution I just committed.

alvarop:
Excellent, thank you!

PR #1200, opened 2024-02-13
Added Extended NEC Protocol to TinyIR

Buzzerb:
I've added the option to receive Extended NEC using TinyIRReceiver with a USE_EXTENDED_NEC_PROTOCOL compile option. It functions similarly to the USE_ONKYO_PROTOCOL flag, mostly reusing its code. 

Although Extended NEC can already be received by using the ONKYO protocol or disabling parity checking, doing so gives up the parity checking which should still be possible on the command part of the transmission. Therefore, I figured a proper implementation was worthwhile.

The README and example file TinyReceiver.ino have been updated to reflect this addition. Tested working with TinyReciever.ino and an UNO rev 3.

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,7 +418,7 @@ http://www.harctoolbox.org/IR-resources.html
 
 
 # Tiny NEC receiver and sender
-For applications only requiring NEC or FAST -see below- protocol, there is a special receiver / sender included,<br/>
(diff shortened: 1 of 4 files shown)


ArminJo:
Awesome! The first PR which also maintains the documentation!!!
Thank you very much 🥇

PR #1202, opened 2024-02-15
Add ExtendedNEC support to TinyIRSender and move NEC2 to arguments

Buzzerb:
Added ExtendedNEC support to TinyIRSender. Also redesigned TinyIRSender's handling of sending NEC2 repeats from a compile time flag to an argument in the send functions for NEC variants, as per discussion #1201. Modified the example script TinySender to support these changes/additions and fix a bug. Finally updated/made consistent some comments across various files and the README.

Has been tested working with two Arduino Uno rev 3's with TinySender and SimpleReciever for all NEC variants. Also checked compile sizes using a modified TinySender script. RAM usage has remained the same, Flash usage has remained the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -447,7 +447,7 @@ void loop() {
 #include "TinyIRSender.hpp"
 
 void setup() {
-  sendNECMinimal(3, 0, 11, 2); // Send address 0 and command 11 on pin 3 with 2 repeats.
+  sendNEC(3, 0, 11, 2); // Send address 0 and command 11 on pin 3 with 2 repeats.
 }
 
 void loop() {}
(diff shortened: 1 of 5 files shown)


Buzzerb:
Just noticed the failed checks, trying to see what caused them now

Buzzerb:
Cannot reproduce the compile errors, any ideas?

ArminJo:
Default specification is not allowed for definitions it is alowed only for declarations.
Keep them in the .h file and omit in *.hpp.

I think it is a good idea, to omit the comment "Although the below compile flags..." and the compile flags in TinySender.ino.
I would also propose to extend your changes and change the example, so that all protocols are sent after each other, making the compile flags obsolete and showing simultaneous using of different send protocols.
I hope that is a bit more understandable for the user.

And thanks for extending the comments 👍 .

Buzzerb:
Fixed issue with default specification in definitions as opposed to declarations. 

I agree that is more user friendly. I have updated TinySender.ino to send using (almost*) all protocols simultaneously. Added a comment that despite the large program size this inevitably ends up creating, users can simply delete or comment out unneeded protocols and the final size will be smaller. *(Not all combinations of NEC2 are demonstrated for brevity).

Buzzerb:
> Thank you very much for your professional contribution!!!

No worries, thanks for your help

PR #1253, opened 2024-07-31
Add some missing F() invocations.

leres:
Also use ... println() instead of println("") to get a blank newline.


ArminJo:
Thanks for contributing and finding the flaws 😀 👍

PR #1254, opened 2024-07-31
Edited 2.x and 4.x Example

luiwagner:
Added `Serial.begin(115200);` to the old 2.x and new 4.x example program, for easier copy pasting.


ArminJo:
Thanks for contributing 👍

Issue #1260, opened 2024-08-16
[Bug]: Bumped version to 4.4.1 works without error if "timerStart(s50usTimer);" is replaced with "timerRestart(s50usTimer);"


mboehmerm:
### The type of board you are using.

ESP32 board - first check https://github.com/crankyoldgit/IRremoteESP8266

### The boards name or FQBN (e.g. esp8266:esp8266:d1_mini:eesz=4M3M,xtal=80) for unspecified board.

Nologo ESP32C3 Super Mini

### The board manager URL used for unspecified board.

https://

### What IDE are you using?

Arduino IDE

### What IR protocol are you using?

NEC

### Pin(s) used for IR-receive, if not default.

Arduino pin number: 10

### Example(s) you have checked while hunting the bug.

- [ ] SimpleReceiver
- [ ] ReceiveDemo
- [ ] SendRawDemo
- [ ] ReceiverTimingAnalysis
- [ ] TinyReceiver
- [ ] TinySender
- [ ] ReceiveAndSend
- [ ] SimpleSender
- [ ] SendDemo
- [ ] SendLGAirConditionerDemo
- [ ] UnitTest
- [ ] Other - please specify below
- [X] I checked, if at least one of the examples was working.

### Example(s) to reproduce the issue.

- [ ] SimpleReceiver
- [X] ReceiveDemo
- [ ] SendRawDemo
- [ ] ReceiverTimingAnalysis
- [ ] TinyReceiver
- [ ] TinySender
- [ ] ReceiveAndSend
- [ ] SimpleSender
- [ ] SendDemo
- [ ] SendLGAirConditionerDemo
- [ ] UnitTest
- [X] Other - please specify below

### The library version you are working with.

- [ ] I use the [latest Arduino library version](../releases) and verified this!
- [ ] I use the latest repo version ([download link](../archive/master.zip)) and verified this!

### What are the steps to reproduce this issue?

I tested the Bumped version to 4.4.1 with Arduino IDE, esp32 Core 3.0.4 and an ESP32-C3 Super Mini.
I used IRremote version 4.4.0 and IRTimer.hpp from the "Bumped version to 4.4.1".

It works for me without error if i replace in IRTimer.hpp
```
timerStart(s50usTimer);" 
```
with
```
timerRestart(s50usTimer);" 
```
Otherwise i get the error message : "E (4118) gptimer: gptimer_start(348): timer is not enabled yet"

Test program
```
[...]

PR #1261, opened 2024-08-20
resolve the Issue #1260

akellai:
avoid error message : "E (4118) gptimer: gptimer_start(348): timer is not enabled yet"


ArminJo:
Thats not the correct solution 😞

PR #1314, opened 2025-08-17
Marantz-RC5 extension support

arjung467:
Added support for Marantz extension of RC5 protocol (sending only). Marantz uses standard RC5x  (13 bit) as well as an own version that adds a pause (4x 889 us) after the address and add adds another word (the "extension") to the signal so the extended message is 19 bit long.

This topic had been in discussion for many years in several threads - so far I haven't found it properly covered in a standard IR library.

What is not included in my patch is an implementation of the decoding feature of IrRemote to also account for the Marantz-specific RC5x implementation. For me this was not of relevance as the Marantz documentation provides the codes for all major devices with Marantz-RC5 control. Unfortunately, older emote controls from Marantz are not available anymore so that I feel it crucial to be able to keep these great HiFi devices from Marantz remote-controllable once the last of my remotes has died ....

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Available as [Arduino library "IRremote"](https://www.arduinolibraries.info/libr
 # Supported IR Protocols
 ` NEC / Onkyo / Apple ` &nbsp; &nbsp; ` Denon / Sharp ` &nbsp; &nbsp; ` Panasonic / Kaseikyo `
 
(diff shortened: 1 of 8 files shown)


arjung467:
tLastBitValue only matters if a 0 is sent (i.e. going through the else branch). As your changes have only an impact in the case of a 1, they don't matter. So both versions will produce exactly the same signals.
Your changes now produce a slight inconsistency in how data starting with a 1 are initialized prior to the loop. Actually, one could always initialize tLastBitValue with a 0 without compromising the result.

PR #1343, opened 2026-02-10
Add OpenLASIR protocol support

danielweidman:
I'm putting together a standard for IR packets that can be used by DIY electronic devices that fit a certain set of use cases. I added support for it in this library because I know it's very popular (I have gotten great use out of it in the past!). It would be great if support could be merged into the main release version so that tinkerers could just pull the main IRremote library from the Arduino IDE instead of having to use my fork. I understand this not likely to gain much widespread use, so I included it in the "exotic" protocol set.

Info about the protocol (IR layer is very similar to NEC Extended, just modified a bit to prevent interference): https://github.com/danielweidman/OpenLASIR

Example code for a laser tag game (using this version of IRremote and [OpenLASIR_Utils.h](https://github.com/danielweidman/OpenLASIR/blob/master/arduino/OpenLASIR_Utils.h)): https://github.com/danielweidman/OpenLASIR/blob/master/examples/arduino/lasertag/lasertag.ino

If it would be helpful, I could put together a more minimal example to put in the main `Arduino-IRremote/examples` folder, but I'm not sure if the demand is there.

Thanks very much for maintaining this excellent resource!

### DIFF
--- a/src/IRReceive.hpp
+++ b/src/IRReceive.hpp
@@ -593,12 +593,6 @@ bool IRrecv::decode() {
     }
 #endif
 
-#if defined(DECODE_OPENLASIR)
-    IR_TRACE_PRINTLN(F("Attempting OpenLASIR decode"));
-    if (decodeOpenLASIR()) {
-        return true;
-    }
-#endif
 
 #if defined(DECODE_PANASONIC) || defined(DECODE_KASEIKYO)
     IR_TRACE_PRINTLN(F("Attempting Panasonic/Kaseikyo decode"));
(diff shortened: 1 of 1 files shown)


danielweidman:
> Thanks for contributing! Great work! Only one recommendation: In IRReceive.hpp can you put the decoding `IR_TRACE_PRINTLN(F("Attempting OpenLASIR decode"));` after the block `#if defined(DECODE_MAGIQUEST) ` ?

Great catch! Thank you. Definitely doesn't make sense to include it as the second protocol in the whole list, haha.

ArminJo:
It would be nice to have an OpenLASIRDemo example and if you could add it to the SendDemo and to the UnitTest examples.
You have to rebase, I committed al my local changes recently.

PR #1344, opened 2026-02-11
Add unit test and example for OpenLASIR protocol

danielweidman:
This PR adds an example OpenLASIR transmission to `SendDemo.ino`, and unit tests to `UnitTest.ino`.

Note about the unit tests (I hope I describe this okay):
In UnitTest.ino, the OpenLASIR packets are interpreted as ONKYO instead of OpenLASIR. This makes sense because (as I understand it) any NEC-timing-based message will get identified at ONKYO if the main NEC/NECext parity check fails and `#define DECODE_ONKYO` is set, because ONKYO is just NEC without the parity check. I intentionally didn't change that behavior (didn't want to risk breaking compatibility for legit ONKYO packets that just happen to match OpenLASER's parity check on address), so OpenLASIR packets will only get correctly identified as OpenLASIR packets if `#define DECODE_ONKYO` is NOT used.

The good news here is that the actual address and command values for the OpenLASIR-but-seen-as-ONKYO packets will be equivalent (OpenLASIR just has a lower max address value), so the unit tests still pass!

If you do not like the fact that the OpenLASIR packets are printed as being ONKYO in the receive part of the unit tests, I could do one of the following:
- Break out OpenLASIR unit tests into a separate file (which would not have `#define DECODE_ONKYO`)
- Update the library to recognize a packet that (passes OpenLASIR parity check but fails the NEC/NECext parity check), as an OpenLASIR packet rather than ONKYO (even if `#define DECODE_ONKYO` is provided)
  - This would risk breaking ONKYO compatibility for some edge cases

### DIFF
--- a/examples/UnitTest/UnitTest.ino
+++ b/examples/UnitTest/UnitTest.ino
@@ -153,16 +153,6 @@ PROGMEM // this crashes on ESP8266
         560/*0000 0x00*/, 560, 1690, 560, 1690, 560, 1690, 560, 560/*1110 Inverted 8 of command*/, 560, 1690, 560, 1690, 560, 1690,
         560, 1690/*1111 inverted 0 of command*/, 560 /*stop bit*/}; // Using exact NEC timing
 
-const uint16_t rawIRTimingsOpenLASIR[]
-#if defined(__AVR__)
-PROGMEM // this crashes on ESP8266
-#endif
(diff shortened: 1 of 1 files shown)


danielweidman:
Looks like my new unit tests took the `UnitTest.ino` size above 32256 bytes and that is breaking the [arduino:avr:uno|DEBUG_TRACE - test compiling examples](https://github.com/Arduino-IRremote/Arduino-IRremote/actions/runs/21891213810/job/63197179579?pr=1344#logs) action. Maybe a sign that I should split it into a separate file? Or else I I'd be totally fine just getting rid of the raw timings unit test I added (that would get rid of the `rawIRTimingsOpenLASIR` array). Let me know!

